PR #3, opened 2021-09-02
Make traits public.

porcuquine:
These modules need to be public so users of `nova` can implement the traits and create circuits.


porcuquine:
Aha, you fixed this while I was working. Thanks.

PR #8, opened 2021-10-12
Add support for using bellperson to generate R1CS.

porcuquine:


Cargo.toml
```diff
@@ -11,6 +11,8 @@ license-file = "LICENSE"
 keywords = ["zkSNARKs", "cryptography", "proofs"]
 
 [dependencies]
+bellperson =  { git = "https://github.com/filecoin-project/bellperson", branch = "nova-minimal" }
```
srinathsetty, line 14: TODO: update the version once the bellperson PR merges.

src/bellperson/shape_cs.rs
```diff
@@ -0,0 +1,328 @@
...
+      (Index::Input(ref a), Index::Input(ref b)) => a.cmp(b),
+      (Index::Aux(ref a), Index::Aux(ref b)) => a.cmp(b),
+      (Index::Input(_), Index::Aux(_)) => Ordering::Less,
+      (Index::Aux(_), Index::Input(_)) => Ordering::Greater,
+    }
+  }
+}
+
+#[allow(clippy::upper_case_acronyms)]
+/// TODO: document
```
srinathsetty, line 48: It would be great to remove all the TODOs

porcuquine: Will do. I thought I had, but apparently not in this file.

src/bellperson/prover.rs
```diff
@@ -0,0 +1,216 @@
+//! Support for generating R1CS witness using bellperson.
```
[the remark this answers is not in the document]
porcuquine: @srinathsetty If you want to rename `ProvingAssignment` in conjunction with this change, can you suggest a good alternative?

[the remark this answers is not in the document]
srinathsetty: @porcuquine is it worth adding a small test case for code in this PR so if we change anything in the future we can use it as a sanity check?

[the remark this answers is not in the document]
srinathsetty: Cool, sounds good! thanks for the pointer!

src/bellperson/shape_cs.rs
porcuquine: Done.

PR #14, opened 2022-02-11
Circuit gadgets: RO gadget based on poseidon + Smul

iontzialla:


porcuquine (commented):
I made a few notes, which may help (understanding that the goal is not an optimized implementation initially).

The one point which is critical is that you need to not create new constants every time you want to perform a Poseidon hash. `neptune` has to allocate all constants and perform a lot of calculations, including sparse-matrix optimizations, when constants are created. This is expected to be done only once per program and for that reason is not even particularly optimized. If you call this on every hash, performance will be shockingly bad.

src/gadgets/ecc_circuit.rs
```diff
@@ -0,0 +1,540 @@
...
+use rand::rngs::OsRng;
+
+#[derive(Clone)]
+pub struct AllocatedPoint<Fp>
+where
+  Fp: PrimeField,
+{
+  x: AllocatedNum<Fp>,
+  y: AllocatedNum<Fp>,
+  is_infinity: AllocatedNum<Fp>, //TODO: Make this Allocatedbit?
```
porcuquine, line 21: Yes, make it an `AllocatedBit`. That will make boolean operations much cheaper (and much less error-prone).

PR #34, opened 2022-04-23
Wasm support

samuelburnham:


Cargo.toml
```diff
@@ -23,7 +23,11 @@ rand_chacha = "0.3"
 itertools = "0.9.0"
 subtle = "2.4"
 pasta_curves = "0.3.0"
-neptune = "6"
+neptune = "6.1"
 generic-array = "0.14.4"
-bellperson-nonnative = "0.2.0"
+bellperson-nonnative = { git = "https://github.com/lurk-lang/bellperson-nonnative", branch = "sb/wasm-support", default-features = false }
```
srinathsetty, line 28: Could we update this to 0.2.1 which includes wasm support?

srinathsetty:
Thanks, @samuelburnham! The PR is marked "draft". Is it ready, or do you need to add additional commits?

The main branch didn't restrict `neptune` so the build was broken with the latest neptune. We made small fixes (which conflicts with your PR, would you be able to resolve them?)

PR #76, opened 2022-05-24
Refactor to allow non-determinisim.

porcuquine:
This PR refactors proving in a number of ways. At least some of these changes are necessary to support determinism in the computation of F. The existing code didn't allow this, since `compute` was always called on the same circuit prototype contained in the public parameters.

- separate circuit/output computation and synthesis into two traits
- at the lowest level, support caller-controlled stepping of the proof
- for convenience, allow caller to provide iterators yielding successive outputs and (un-synthesized) circuits
- most simply, and consistent with current implementation, automatically create such iterators when ComputeStep is implemented

This change also makes `num_steps` optional, and the state accumulated in a `RecursiveSNARK` tracks the actual number of steps used. This allows for applications where the number of steps is not known in advance.

It's likely that further small changes may be needed when integrating downstream consumers who will need this functionality, but until integration happens, it is difficult to anticipate what that may entail beyond my best guesses here.

NOTE: in the future, we should support a recursive proof which does not expose the number of steps. This will be important for applications in which knowing the number of steps required to compute a final result leaks information. Having the option to generate proofs without revealing the number of steps will eliminate effective  'timing attacks' on such applications.

---

UPDATE: I added support for non-unary functions, building on the first part. I know that makes this a fairly large PR, but I had to rework a fair amount of the foundation to get here.

The basic idea is that for higher-arity functions, the implementer should implement methods (`compute_inner` and `synthesize_step_inner`, which operate on and return a vector of scalars. The goal was to make the actual mechanism as hidden from the caller/user as possible, but some leakage was unavoidable so far.

Under the hood, at each step, the inputs are hashed to get the true unary input for the step function, F — and the outputs are again hashed to produce the unary output. Both hashes are proved in the circuit.

Because we need to avoid storing the initial input (Z0) in the public parameters, it's necessary to first create an 'empty' input used to generate the R1CS shape. That means the caller *does* (unfortunately) have to create the poseidon constants for the relevant scalar and arity. This is not especially difficult, but it would be nicer if it could have been avoided. This wart means that the constants are generated twice in this case. With more machinery, the public parameters and IO enum could be more symbiotic, but for now I think this is a reasonable tradeoff.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -95,6 +95,12 @@ where
       c_primary.clone(),
       ro_consts_circuit_primary.clone(),
     );
+    let a1 = if A1::to_usize() == 1 {
+      None
+    } else {
+      circuit_primary.arity_consts()
+    };
+
     let mut cs: ShapeCS<G1> = ShapeCS::new();
     let _ = circuit_primary.synthesize(&mut cs);
     let (r1cs_shape_primary, r1cs_gens_primary) = (cs.r1cs_shape(), cs.r1cs_gens());
@@ -106,20 +112,15 @@ where
       c_secondary.clone(),
       ro_consts_circuit_secondary.clone(),
     );
-    let mut cs: ShapeCS<G2> = ShapeCS::new();
-    let _ = circuit_secondary.synthesize(&mut cs);
-    let (r1cs_shape_secondary, r1cs_gens_secondary) = (cs.r1cs_shape(), cs.r1cs_gens());
-
-    let a1 = if A1::to_usize() == 1 {
-      None
-    } else {
-      Some(PoseidonConstants::<<G1 as Group>::Scalar, A1>::new())
-    };
     let a2 = if A2::to_usize() == 1 {
       None
     } else {
-      Some(PoseidonConstants::<<G2 as Group>::Scalar, A2>::new())
+      circuit_secondary.arity_consts()
     };
+
+    let mut cs: ShapeCS<G2> = ShapeCS::new();
+    let _ = circuit_secondary.synthesize(&mut cs);
+    let (r1cs_shape_secondary, r1cs_gens_secondary) = (cs.r1cs_shape(), cs.r1cs_gens());
 
     Self {
       ro_consts_primary,
@@ -726,15 +727,11 @@ mod tests {
   #[derive(Clone)]
   struct CubicIncrementingCircuit<'a, F: PrimeField> {
     z_n: IO<'a, F, U2>,
-    _p: PhantomData<F>,
   }
 
   impl<'a, F: PrimeField> CubicIncrementingCircuit<'a, F> {
     fn new(z_n: IO<'a, F, U2>) -> Self {
-      Self {
-        z_n,
-        _p: Default::default(),
-      }
+      Self { z_n }
     }
   }
 
@@ -797,10 +794,7 @@ mod tests {
       let y_next = y + F::one();
       let z_vec_next = vec![x_next, y_next];
       let z_n = IO::Vals(z_vec_next.clone(), p);
-      let next = Self {
-        z_n,
-        _p: Default::default(),
-      };
+      let next = Self { z_n };
 
       Some((next, z_vec_next))
     }
@@ -977,12 +971,10 @@ mod tests {
   #[test]
   fn test_ivc_binary() {
     // produce public parameters
-    let p = PoseidonConstants::<S2, U2>::new();
-
     let pp =
       PublicParams::<G1, G2, TrivialTestCircuit<S1>, CubicIncrementingCircuit<S2>, U1, U2>::setup(
         TrivialTestCircuit::default(),
-        CubicIncrementingCircuit::new(IO::Empty(&p)),
+        CubicIncrementingCircuit::new(IO::Blank),
       );
 
     let num_steps = 1;
@@ -1013,7 +1005,6 @@ mod tests {
     for _i in 0..num_steps {
       zn_secondary_direct = CubicIncrementingCircuit {
         z_n: zn_secondary_direct.clone(),
-        _p: Default::default(),
       }
       .compute_io(&zn_secondary_direct)
       .unwrap()
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -32,7 +32,7 @@ use r1cs::{
   R1CSGens, R1CSInstance, R1CSShape, R1CSWitness, RelaxedR1CSInstance, RelaxedR1CSWitness,
 };
 use traits::{
-  AbsorbInROTrait, ComputeStep, Group, HashFuncConstantsTrait, HashFuncTrait, StepCircuit,
+  AbsorbInROTrait, StepCompute, Group, HashFuncConstantsTrait, HashFuncTrait, StepCircuit,
 };
 
 type ROConstants<G> =
@@ -120,12 +120,12 @@ where
 }
 
 /// State of iteration over ComputeStep
-pub struct ComputeState<S: ComputeStep<F> + StepCircuit<F>, F: PrimeField> {
+pub struct ComputeState<S: StepCompute<F> + StepCircuit<F>, F: PrimeField> {
   circuit: S,
   val: F,
 }
 
-impl<F: PrimeField, S: ComputeStep<F> + StepCircuit<F> + Clone> Iterator for ComputeState<S, F> {
+impl<F: PrimeField, S: StepCompute<F> + StepCircuit<F> + Clone> Iterator for ComputeState<S, F> {
   type Item = (S, F);
 
   fn next(&mut self) -> Option<Self::Item> {
@@ -248,7 +248,7 @@ where
     Ok(state)
   }
 
-  /// create a new `RecursiveSNARK` for circuits implementing ComputeStep
+  /// create a new `RecursiveSNARK` for circuits implementing StepCompute
   pub fn prove(
     pp: &PublicParams<G1, G2, C1, C2>,
     num_steps: Option<usize>,
@@ -256,8 +256,8 @@ where
     z0_secondary: G2::Scalar,
   ) -> Result<Self, NovaError>
   where
-    C1: ComputeStep<G1::Scalar>,
-    C2: ComputeStep<G2::Scalar>,
+    C1: StepCompute<G1::Scalar>,
+    C2: StepCompute<G2::Scalar>,
   {
     if let Some(num_steps) = num_steps {
       let mut primary_iterator = ComputeState {
@@ -576,7 +576,7 @@ mod tests {
     }
   }
 
-  impl<F> ComputeStep<F> for TrivialTestCircuit<F>
+  impl<F> StepCompute<F> for TrivialTestCircuit<F>
   where
     F: PrimeField,
   {
@@ -626,7 +626,7 @@ mod tests {
     }
   }
 
-  impl<F> ComputeStep<F> for CubicCircuit<F>
+  impl<F> StepCompute<F> for CubicCircuit<F>
   where
     F: PrimeField,
   {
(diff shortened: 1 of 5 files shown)


src/traits.rs
```diff
@@ -134,18 +134,21 @@ impl<T, Rhs, Output> ScalarMul<Rhs, Output> for T where T: Mul<Rhs, Output = Out
...
     &self,
     cs: &mut CS,
     z: AllocatedNum<F>,
   ) -> Result<AllocatedNum<F>, SynthesisError>;
+}
 
-  /// Execute the circuit for a computation step and return output
-  fn compute(&self, z: &F) -> F;
+/// A helper trait for computing a step of the incremental computation (i.e., F itself)
+pub trait ComputeStep<F: PrimeField>: Sized {
```
srinathsetty, line 149: Nit: Does it make sense to call this `StepNative` or `StepNonCircuit` to mirror with `StepCircuit`?

src/lib.rs
```diff
@@ -526,13 +691,15 @@ mod tests {
...
       &pp,
-      num_steps,
+      Some(num_steps),
       <G1 as Group>::Scalar::one(),
       <G2 as Group>::Scalar::zero(),
     );
     assert!(res.is_ok());
     let recursive_snark = res.unwrap();
 
+    assert_eq!(num_steps, recursive_snark.num_steps);
```
srinathsetty, line 701: Is this assertion a sanity check or required for verifying that the prover faithfully executed the specified number of steps? If it is the latter, shouldn't this automatically be checked by the verify method?

src/lib.rs
```diff
@@ -579,13 +747,15 @@ mod tests {
...
```
srinathsetty, line 757: Same comment as above.

srinathsetty:
Thanks, @porcuquine for the PR! I left some preliminary comments.

Given the focus is to allow non-determinism, could we add/extend one of the tests to include some non-deterministic step function (as a sanity check and to demonstrate the use of public APIs)?

src/traits.rs
porcuquine: I think 'compute' is pretty accurate, but I pushed a commit changing to `StepCompute` more closely rhyming with `StepCircuit`.

src/lib.rs
porcuquine: It was a sanity check, but you are right that since it is now included in `RecursiveSNARK`, we should fail verification if it doesn't match what is supplied. I added that logic to the verifier with some minimal negatives tests that `num_steps` must be correct.

porcuquine:
Okay, I added a test exercising the new procedure by precomputing nondeterministic hints and storing them in the circuit. I needed to tweak the implementation a little to make this work (so the test was especially useful!). This also revealed that the concrete circuits never needed to stored in the RecursiveSNARK, so I removed.

src/poseidon.rs
```diff
@@ -36,15 +36,35 @@ where
   /// Generate Poseidon constants for the arities that Nova uses
   #[allow(clippy::new_without_default)]
   fn new() -> Self {
-    let constants27 = PoseidonConstants::<Scalar, U27>::new_with_strength(Strength::Strengthened);
-    let constants32 = PoseidonConstants::<Scalar, U32>::new_with_strength(Strength::Strengthened);
+    let constants27 = PoseidonConstants::<Scalar, U27>::new_with_strength(Strength::Standard);
+    let constants32 = PoseidonConstants::<Scalar, U32>::new_with_strength(Strength::Standard);
```
porcuquine, line 40: Standard is fine and will save some constraints.

src/lib.rs
```diff
@@ -461,143 +720,388 @@ mod tests {
...
-      G1,
-      G2,
-      TrivialTestCircuit<<G1 as Group>::Scalar>,
-      CubicCircuit<<G2 as Group>::Scalar>,
-    >::setup(
-      TrivialTestCircuit {
-        _p: Default::default(),
-      },
-      CubicCircuit {
+    let p = PoseidonConstants::<S2, U2>::new();
```
porcuquine, line 988: Here is an example of the caller/user needing to explicitly allocate PoseidonConstants.

src/lib.rs
```diff
@@ -461,143 +720,388 @@ mod tests {
...
+
+    let pp =
+      PublicParams::<G1, G2, TrivialTestCircuit<S1>, CubicIncrementingCircuit<S2>, U1, U2>::setup(
+        TrivialTestCircuit::default(),
+        CubicIncrementingCircuit::new(IO::Empty(&p)),
```
porcuquine, line 993: The PoseidonConstants are used here, in the blank/empty circuit.

src/lib.rs
```diff
@@ -101,6 +110,17 @@ where
     let _ = circuit_secondary.synthesize(&mut cs);
     let (r1cs_shape_secondary, r1cs_gens_secondary) = (cs.r1cs_shape(), cs.r1cs_gens());
 
+    let a1 = if A1::to_usize() == 1 {
+      None
+    } else {
+      Some(PoseidonConstants::<<G1 as Group>::Scalar, A1>::new())
```
srinathsetty, line 116: Do we need to check if the arity is less than a certain number? Or is the call to PoseidonConstants always going to succeed?

src/lib.rs
```diff
@@ -138,26 +223,30 @@ where
...
-    pp: &PublicParams<G1, G2, C1, C2>,
-    num_steps: usize,
+  /// Initialize a `RecursiveSNARK`.
+  pub fn init(
+    pp: &PublicParams<G1, G2, C1, C2, A1, A2>,
     z0_primary: G1::Scalar,
     z0_secondary: G2::Scalar,
+    z1_primary: G1::Scalar,
+    z1_secondary: G2::Scalar,
+    first_primary_circuit: &C1,
```
srinathsetty, line 247: Change to `first_circuit_primary` and `first_circuit_secondary` to rhyme with the naming style of other arguments?

src/lib.rs
```diff
@@ -138,26 +223,30 @@ where
...
+    first_secondary_circuit: &C2,
   ) -> Result<Self, NovaError> {
-    if num_steps == 0 {
```
srinathsetty, line 157: How do we handle this case without this code in the new version?

src/lib.rs
```diff
@@ -191,145 +280,296 @@ where
...
+        z0_primary,
+        z0_secondary,
+        &mut primary_iterator,
+        &mut secondary_iterator,
+      )
+    }
+  }
+
+  /// create a new `RecursiveSNARK` from primary and secondary iterators
+  pub fn prove_with_iterators(
```
srinathsetty, line 387: Would this need to be public?

src/lib.rs
```diff
@@ -53,18 +58,22 @@ where
   ro_consts_circuit_secondary: ROConstantsCircuit<<G1 as Group>::Base>,
   r1cs_gens_secondary: R1CSGens<G2>,
   r1cs_shape_secondary: R1CSShape<G2>,
-  c_primary: C1,
-  c_secondary: C2,
   params_primary: NIFSVerifierCircuitParams,
   params_secondary: NIFSVerifierCircuitParams,
+  c_primary: C1,
+  c_secondary: C2,
+  a1: Option<PoseidonConstants<<G1 as Group>::Scalar, A1>>,
```
srinathsetty, line 65: I think we should be using `ROConstants` instead of `PoseidonConstants` (there was a TODO at the top of the file to completely avoid including poseidon.rs from lib.rs). In the short term, if the goal is to support a few arities of interest (e.g., 2 for VDFs and 6 for Lurk), could we just add those aritites to the hard-coded constants that we generate? This will likely avoid a lot of complexity. What do you think?

srinathsetty: We could then make implementations of `ROConstants` be able to handle arbitrary arities via the sponge construction.

src/lib.rs
```diff
@@ -138,26 +223,30 @@ where
   zn_secondary: G2::Scalar,
   _p_c1: PhantomData<C1>,
   _p_c2: PhantomData<C2>,
+  a1: PhantomData<A1>,
```
srinathsetty, line 226: Should these be `_p_a1` and `_p_a2`?

src/lib.rs
porcuquine: It won't fail at runtime. If you try to specify a concrete array size that doesn't have an implementation of Arity in Neptune (maybe > 36?) then it will fail at compile time. So I don't think a runtime check is required.

src/lib.rs
porcuquine: Changed.

src/lib.rs
porcuquine: The error will be returned if an iterator has no first item:
```
https://github.com/microsoft/Nova/pull/76/files/3986df4f2c580d49f449f84ab84c02eb0db1668b#diff-b1a35a68f14e696205874893c07fd24fdb88882b47c23cc0e0c80a30c7d53759R394-R401
```

src/lib.rs
porcuquine: Changed.

src/lib.rs
```diff
@@ -191,145 +280,296 @@ where
...
+      )?
+      .is_some()
+    {
+      i += 1;
+    }
+    Ok(state)
+  }
+
+  /// Prove one step of an iterative computation, mutating the RecursiveSNARK
+  pub fn prove_step(
```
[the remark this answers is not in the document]
porcuquine: Changed.

src/lib.rs
porcuquine: However, I've changed this one back. See answer to question below.

src/traits.rs
```diff
@@ -134,18 +135,152 @@ impl<T, Rhs, Output> ScalarMul<Rhs, Output> for T where T: Mul<Rhs, Output = Out
...
+  pub fn new_vals(&self, vals: Vec<F>) -> Self {
+    match self {
+      IO::Val(_) => unreachable!(),
+      IO::Vals(_, p) => IO::Vals(vals, p),
+      IO::Empty(_) => unreachable!(),
+    }
+  }
+}
+
+/// A helper trait for computing a step of the incremental computation (i.e., F itself)
```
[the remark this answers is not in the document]
porcuquine: Yes, a caller might want more control, in which case they can implement Iterator directly. This will bypass the iterator automatically created using `ComputeStep`. That is also why `prove_with_iterators` is public. When the caller implements an iterator, they can bypass `prove` and call the lower-level function.

src/lib.rs
porcuquine: These fields are actually used to simplify initializing `IO::Vals` later.

src/lib.rs
porcuquine: I agree with you on the *general* topology, and I have pushed a commit which I think finally addresses the issues. I moved the arity poseidon constants into the verifier circuit. I didn't put them in `ROConstants` directly for several related reasons:

1. If we retain generality, that would require putting an `Arity` type parameter everywhere, including places it didn't belong.
2. If we discard generality, it would actually have added a lot of complexity and cost around 'dynamic allocation' of specific constants.
3. They're not *really* part of the RO anyway. As implemented here, they're a helper for the application function F.

In any case, my recent commit now successfully hides the PoseidonConstants from the caller almost completely. In the case that a caller wants to implement the `ComputeStep` trait for a non-unary function, they need to handle (but not generate) constants in the `compute_inner` method. This can probably also be eliminated, but I think it's livable for now.

porcuquine:
@srinathsetty I have resolved the conflicts with `main`, which involved adding arity support for `CompressedSNARK`.

porcuquine:
I have pushed a few more small changes, including a small refactor allowing for cleaner trait implementations. The first-order goal of this PR is now complete; the interfaces are now sufficient to use Nova recursion in the VDF proof as implemented here: https://github.com/protocol/vdf/pull/14.

porcuquine:
I'm closing this, since its purpose is and/or will be fulfilled by other work now.

PR #92, opened 2022-07-15
ecdsa signature proof

arthurgreef:


examples/ecdsa/circuit.rs
```diff
@@ -0,0 +1,322 @@
...
+      x,
+      y,
+      is_infinity: false,
+    }
+  }
+}
+
+// An ECDSA signature
+#[derive(Clone, Debug)]
+pub struct EcdsaSignature<Fb, Fs>
```
srinathsetty, line 44: @arthurgreef Thanks again for the example! I pushed a commit to simplify (by using the library-provided trivial circuit).

I had some questions about this code. My understanding is that an ECDSA signature typically consists of (r, s). So besides (pk, r, s), what are the other things here? I'm thinking `g` is the generator of the group (perhaps Nova's Group definition should be enhanced so it returns the generator of the group? would that make the circuit call the trait method to get the generator so it is hard-coded into the circuit rather than provided as input?) Also, what is the purpose of `c`?

arthurgreef: @srinathsetty A signature is (r,s) and signature verification requires (H(message), pk, signature). (c)ommitment = H(message). In the example, I used a small message so that according to [bellman it could be used without hashing](https://github.com/filecoin-project/bellperson-gadgets/blob/main/src/eddsa.rs#:~:text=//%20In%20a%20VERY,directly%20without%20hashing). The group definition can be enhanced so that the generator (g) is hard-coded into the circuit.

srinathsetty: @arthurgreef thanks! I added some comments in the code to capture this thread. I also removed some allocated variables (please let me know if that is okay or if I introduced an error).

I added a note at the beginning to say the example can be adapted to prove the validity of a certificate chain (where step i proves that the knowledge of a valid signature on step i's public key using step (i-1)'s public key and the public key of step 0 is in the public IO). I think it is only a small change and might make it more interesting and non-trivial.

arthurgreef: Excellent. Thanks.

PR #96, opened 2022-07-21
key generation proof

arthurgreef:
Algorithm taken from [Proof of private key ownership](https://medium.com/zokrates/building-identity-linked-zksnarks-with-zokrates-a36085cdd40). Zokrates has the notation of private input. Is the input private by default for Nova?

### DIFF
--- a/examples/keygen.rs
+++ b/examples/keygen.rs
@@ -1,0 +1,425 @@
+//! Demonstrates how to use Nova to produce a recursive proof of the correct execution of
+//! iterations of the key generation function.
+//! Generate public key using G1 curve, and prove using G2 curve.
+
+use bellperson::{
+  gadgets::{boolean::AllocatedBit, num::AllocatedNum},
+  ConstraintSystem, SynthesisError,
+};
+use ff::{Field, PrimeField, PrimeFieldBits};
+use generic_array::typenum::U5;
+use neptune::{
+  circuit::poseidon_hash,
+  poseidon::{Poseidon, PoseidonConstants},
+  Strength,
+};
+use nova_snark::{
+  gadgets::ecc::AllocatedPoint,
+  traits::{
+    circuit::{StepCircuit, TrivialTestCircuit},
+    Group as Nova_Group,
+  },
+  CompressedSNARK, PublicParams, RecursiveSNARK,
+};
+use pasta_curves::{
+  arithmetic::CurveAffine,
+  group::{Curve, Group},
+};
+use rand::{rngs::OsRng, RngCore};
+use subtle::Choice;
+
+type G1 = pasta_curves::pallas::Point;
+type G2 = pasta_curves::vesta::Point;
+type S1 = nova_snark::spartan_with_ipa_pc::RelaxedR1CSSNARK<G2>;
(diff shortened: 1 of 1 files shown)


examples/keygen.rs
```diff
@@ -0,0 +1,425 @@
...
+    let point = G1::generator() * s.0;
+    Self(point)
+  }
+}
+
+// An affine point coordinate that is on the curve.
+#[derive(Clone, Copy, Debug)]
+pub struct Coordinate<F>
+where
+  F: PrimeField<Repr = [u8; 32]>,
```
srinathsetty, line 60: Is it possible to avoid placing restrictions on Repr? I was able to drop such restrictions in the signature example, please see `examples/signature.rs`.

examples/keygen.rs
```diff
@@ -0,0 +1,425 @@
...
+  }
+}
+
+// An key generation proof that we will use on the primary curve
+#[derive(Clone, Debug)]
+pub struct KeyGenerationCircuit<F>
+where
+  F: PrimeField<Repr = [u8; 32]>,
+{
+  pub z_g: Coordinate<F>,
```
srinathsetty, line 109: I was a bit confused about the use of `z_g`, `z_pk`, and `z_sk`. Are these needed?

examples/keygen.rs
```diff
@@ -0,0 +1,425 @@
+//! Demonstrates how to use Nova to produce a recursive proof of the correct execution of
+//! iterations of the key generation function.
```
srinathsetty, line 2: I had a basic question about this example: What exactly is this proving? It seems like it is proving the knowledge of a set of secret keys? Does it make sense to turn this into proving knowledge of (one or more) secret keys such that their public keys are *in* a set of known public keys? Without such a restriction, it seems like the prover can just make up a set of secret keys and prove their knowledge, right?

arthurgreef: The use case for this example was taken from [Proof of private key ownership](https://medium.com/zokrates/building-identity-linked-zksnarks-with-zokrates-a36085cdd40#:~:text=the%20zkSNARK!-,Proof%20of%20Private%20Key%20Ownership,-In%20the%20context). Are you describing a membership proof? Changing the example to a better one is fine with me.

srinathsetty: Yes, the example is fine. But is the example making the claimed public key a public output, so the verifier can check if it is the same as some designated/well-known public key? If not, the proven statement isn't very useful, right? Because proving the knowledge of *some* secret key that corresponds to *some* public key isn't hard, right?

arthurgreef: From the web page the example is a proof of identity. The public key is a public output that the verifier would check against a public key that could be retrieved from a DID document for example. It may be the case that the example in isolation is not useful. The referenced web page for example uses two proofs for an identity-linked proof of knowledge.

- we have one function which proves knowledge of a secret
- we have another function which checks that a given public key was derived from a given private key

srinathsetty:
Closing this as Nova's interfaces have diverged substantially from what is used here. Please reopen with an updated PR.

PR #99, opened 2022-07-22
chain link proof example

arthurgreef:
Demonstrates how an attribute equality proof can be used with a chain function to prove the veracity of a provenance chain of linked data.


srinathsetty:
Closing this as Nova's interfaces have diverged substantially from what is used here. Please reopen with an updated PR.

PR #47, opened 2022-04-30
Serde serialization

arthurgreef:
Draft pull request as this pull request depends on a pull request on the pasta_curves repo: https://github.com/zcash/pasta_curves/pull/36/commits

Fixes #46

PR #123, opened 2023-01-18
Add serde proof serialization

samuelburnham:
This PR derives serde for all the types needed for Lurk proofs, based on #47.

In order to serialize the pasta curves, two forks are needed for the time being:
- `fil_pasta_curves`, which Nova already depends on via Neptune
- `lurk-pasta-msm`, which pulls in `fil_pasta_curves`

While serde serialization is already implemented in the latest version of `pasta_curves`, it also has significant breaking changes such as new versions of the `ff` and `group` crates. Rather than manually updating each crate in the dependency tree, Neptune and Lurk have elected to use a fork that only adds the serde changes for now, as explained in https://github.com/filecoin-project/rust-fil-proofs/issues/1653.


samuelburnham:
@microsoft-github-policy-service agree [company="{Yatima Inc}"]

srinathsetty:
@samuelburnham Thanks for the PR! This PR has conflicts due to the Wasm and related PRs. Would it be possible to resolve conflicts?

PR #125, opened 2023-01-18
Add Wasm support

samuelburnham:
Enables compiling to `wasm32-unknown-unknown`, which requires conditionally compiling `pasta-msm` since it doesn't support Wasm.

If the `rand` crate is desired for ease of use I can add it back in, but would have to conditionally compile `getrandom` as shown in https://docs.rs/getrandom/latest/getrandom/#webassembly-support


src/pasta.rs
```diff
@@ -58,6 +58,18 @@ impl Group for pallas::Point {
     }
   }
 
+  #[cfg(target_arch = "wasm32")]
```
srinathsetty, line 61: This makes the proof generation slower under wasm by using a naive MSM implementation. Is pasta-msm not compatible with Wasm?

samuelburnham: I don't believe it is, at least not officially, as I haven't been able to get it to compile. I opened an issue to ask them https://github.com/supranational/pasta-msm/issues/1, but in the meantime I'd be inclined to merge now and optimize later as Wasm support matures in general.

nalinbhardwaj: Hi! I've just noticed this PR as I was working on adding WASM support to [Nova Scotia](https://github.com/nalinbhardwaj/Nova-Scotia). While its not very thoroughly benchmarked, I've had a patch for Nova with fast multiexp (forking the MSM code from halo2) here: https://github.com/nalinbhardwaj/Nova/tree/wasm-upstream

I can make a PR if you'd prefer that patch to be merged instead!

srinathsetty: @samuelburnham got it!

@nalinbhardwaj sounds exciting! It looks like a serial implementation. Is there a possibility to make it parallel by using rayon? Also, it looks like the multiexp code is general, so is there a possibility to put it in Nova's group trait directly?

nalinbhardwaj: I've just made https://github.com/microsoft/Nova/pull/126 with the MSM changes (and confirmed they work in WASM and speed up small MSMs on x86 as well)

> Also, it looks like the multiexp code is general, so is there a possibility to put it in Nova's group trait directly?

My understanding is that the code is specifically optimised to <=256 bit fields (such as Pasta), so I think it's better not to make it a member of a general Group trait. What do you think?

srinathsetty: @samuelburnham could you please rebase this branch with the latest main?

PR #126, opened 2023-01-25
Speed up MSMs for non-GPU accelerated MSMs and architectures that don't support GPU/semolina

nalinbhardwaj:
Adds fast, parallelisable MSM for compile targets that aren't supported by pasta_msm/semolina. Also uses fast MSM on small MSMs (smaller than 128 size) for _all_ architecture since I'm on my laptop a size 128 MSM is ~3x faster with this code vs. the old non Pippengers code.

The multiexp code is taken from https://github.com/zcash/halo2/blob/main/halo2_proofs/src/arithmetic.rs, all credits to authors there!

Tested end to end with this branch of Nova Scotia: https://github.com/nalinbhardwaj/Nova-Scotia/tree/browser

Follow up from discussion here: https://github.com/microsoft/Nova/pull/125#discussion_r1087001908

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -49,5 +49,5 @@ harness = false
 [features]
 default = [ "bellperson/default", "neptune/default" ]
 
-[target.'cfg(not(target_family = "wasm"))'.dependencies]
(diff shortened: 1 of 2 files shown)


nalinbhardwaj:

PR #129, opened 2023-01-27
parallelize parameter generation

srinathsetty:


src/pasta.rs
```diff
@@ -204,14 +205,20 @@ impl Group for pallas::Point {
...
-      gens.push(hash(&uniform_bytes).to_affine());
+      uniform_bytes_vec.push(uniform_bytes);
     }
-    gens
+    (0..n)
+      .collect::<Vec<usize>>()
+      .into_par_iter()
+      .map(|i| {
+        let hash = Ep::hash_to_curve("from_uniform_bytes");
+        hash(&uniform_bytes_vec[i]).to_affine()
```
nalinbhardwaj, line 219: Was just looking at doing this myself since the [zator](https://github.com/lyronctk/zator) folks mentioned slow performance here:

One other minor optimisation you can make here is to use [`batch_normalize`](https://docs.rs/pasta_curves/latest/pasta_curves/struct.Ep.html#method.batch_normalize) for this conversion. One clean way to do both optimisations is to do the rayon parallelisation manually by splicing the list into `num_threads` pieces

PR #141, opened 2023-02-12
Feature to enable Neptune GPU Poseidon

nalinbhardwaj:
Adds features `cuda` and `opencl` to toggle Neptune to compile and use GPU kernels. On https://github.com/nalinbhardwaj/Nova-Scotia/pull/6 this is a ~10% speedup on a larger circuit on my 2018 MacBook (so likely significant on better machines/GPU as well).

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -30,5 +30,5 @@ jobs:
     - name: Check Rustfmt Code Style
       run: cargo fmt --all -- --check
     - name: Check clippy warnings
-      run: cargo clippy --all-targets --all-features -- -D warnings
+      run: cargo clippy --all-targets -- -D warnings
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -51,3 +51,5 @@ harness = false
 
 [features]
 default = []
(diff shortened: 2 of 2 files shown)


.github/workflows/rust.yml
```diff
@@ -2,33 +2,32 @@ name: Build and Test Nova
...
+    branches: [main]
   pull_request:
-    branches: [ main ]
+    branches: [main]
 
 jobs:
   build:
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v2
```
srinathsetty, line 13: @nalinbhardwaj thanks for the PR! 

What is the change here? It appears to be mostly a formatting change, except for the clippy line. Could we make the diff reflect the actual change?

nalinbhardwaj: Oops, my editor auto linted the file, undid the lint. The change here is to remove `--all-features` (since [GitHub Workflows doesn't support GPU CUDA/OpenCL](https://github.com/github/roadmap/issues/505) the tests are expected to fail for those features)

PR #162, opened 2023-04-18
chore: update to ff/group 0.13 and associated dependencies

huitseeker:
Updates:
- zkcrypto/ff, zkcrypto/group to 0.13,
- bellperson to 0.25,
- neptune to 9.0,
- pasta_curves to 0.5.1, and removes the [fil_pasta_curves fork](https://github.com/lurk-lab/pasta_curves)
- pasta-msm should no longer need a fork (WIP)

Adapts source in function, mostly for const usage and API updates.

## TODOs
- [x] run test on `EC_GPU_FRAMEWORK=cuda`
- [x] run tests on `EC_GPU_FRAMEWORK=opencl` 
- [x] get [sister PR to pasta-msm](https://github.com/supranational/pasta-msm/pull/3) merged

PR #165, opened 2023-04-28
Make ecc tests generic wrt curves

leonardoalt:
This PR makes the ECC tests generic wrt to the curves, as long as a curve implements `nova::traits::Group`, `pasta_curves::arithmetic::CurveAffine` (which is the case for the curves in `halo2curves`, eg), and the `Base` and `Scalar` fields are shared between the traits.

After https://github.com/microsoft/Nova/pull/162 is merged, we can implement other curve cycles and use them directly in the tests such as in https://github.com/microsoft/Nova/commit/8e70bfca67f77bcfd6de5261d76a52fefdcfa951.

### DIFF
--- a/src/gadgets/ecc.rs
+++ b/src/gadgets/ecc.rs
@@ -755,11 +755,8 @@ mod tests {
     {shape_cs::ShapeCS, solver::SatisfyingAssignment},
   };
   use ff::{Field, PrimeFieldBits};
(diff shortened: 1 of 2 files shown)


leonardoalt:
Just rebased on top of `main` after https://github.com/microsoft/Nova/pull/166 got merged.

PR #171, opened 2023-05-22
[test-only] More genericity in tests

huitseeker:
Makes ~bellperson, ivc, circuit, nifs~ all tests generic in the type of the involved curves. 
This PR is a descendent of #165 should greatly help testing Nova w/ another curve cycle. (/cc @leonardoalt)

This shows a bit of (acceptable) non-genericity in the ivc tests, in that we rely on the `CommitmentKeyExt` trait being implemented to allow usage of the IPA as a PC. I'm not sure this is inherently necessary, but perhaps we can lift this restriction in another PR.

### DIFF
--- a/src/nifs.rs
+++ b/src/nifs.rs
@@ -162,10 +162,9 @@ mod tests {
     Ok(())
   }
 
-  fn test_tiny_r1cs_bellperson_with<S, G>()
+  fn test_tiny_r1cs_bellperson_with<G>()
   where
-    S: PrimeField,
-    G: Group<Scalar = S>,
+    G: Group,
   {
     use crate::bellperson::{
       r1cs::{NovaShape, NovaWitness},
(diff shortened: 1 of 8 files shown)


srinathsetty:
Thanks @huitseeker! Is this PR ready, or do you have any more pending updates?

huitseeker:
@srinathsetty it is ready!

src/nifs.rs
```diff
@@ -162,8 +162,11 @@ mod tests {
     Ok(())
   }
 
-  #[test]
-  fn test_tiny_r1cs_bellperson() {
+  fn test_tiny_r1cs_bellperson_with<S, G>()
+  where
+    S: PrimeField,
```
srinathsetty, line 167: @huitseeker A minor nit. Is it necessary for this to be generic over S, when it is already generic over G?

src/nifs.rs
```diff
@@ -206,8 +209,13 @@ mod tests {
...
   }
 
+  #[test]
+  fn test_tiny_r1cs_bellperson() {
+    test_tiny_r1cs_bellperson_with::<S, G>();
+  }
+
   #[allow(clippy::too_many_arguments)]
-  fn execute_sequence(
+  fn execute_sequence<S, G>(
```
srinathsetty, line 218: Same question as above. Is `S` necessary?

src/nifs.rs
huitseeker: No, you're right, here and below, I was re-using the defined alias for `S`, but that's a bit lazy. Will fix.

huitseeker:
@srinathsetty Thanks for the review, I've done an additional pass. I think I've caught:
- making all curve-dependent tests that could be made so generic in the curve groups,
- simplifying the generic parameters involved to their simplest expression.
Please let me know if you spot something I missed.

PR #181, opened 2023-06-09
Add Grumpkin cycle implementation

huitseeker:
This is joint work with @leonardoalt !

This PR pulls in an implementation of, and runs all Nova tests on the [Ethereum-friendly Grumpkin curve cycle](https://hackmd.io/@aztec-network/ByzgNxBfd#2-Grumpkin---A-curve-on-top-of-BN-254-for-SNARK-efficient-group-operations). 

The implementation ~~is~~ was  from the halo2curves repo with the following pending PR thrown in:
https://github.com/privacy-scaling-explorations/halo2curves/pull/47 - ~~I'll update the present PR when that resolves.~~

### DIFF
--- a/src/provider/bn254_grumpkin.rs
+++ b/src/provider/bn254_grumpkin.rs
@@ -1,0 +1,338 @@
+//! This module implements the Nova traits for bn256::Point, bn256::Scalar, grumpkin::Point, grumpkin::Scalar.
+use crate::{
+  provider::{
+    keccak::Keccak256Transcript,
+    pedersen::CommitmentEngine,
+    poseidon::{PoseidonRO, PoseidonROCircuit},
+  },
+  traits::{CompressedGroup, Group, PrimeFieldExt, TranscriptReprTrait},
+};
+use digest::{ExtendableOutput, Input};
+use ff::{FromUniformBytes, PrimeField};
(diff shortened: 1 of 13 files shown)


src/provider/bn254_grumpkin.rs
```diff
@@ -0,0 +1,367 @@
...
+
+use halo2curves::bn256::{
+  G1Affine as Bn256Affine, G1Compressed as Bn256Compressed, G1 as Bn256Point,
+};
+use halo2curves::grumpkin::{
+  G1Affine as GrumpkinAffine, G1Compressed as GrumpkinCompressed, G1 as GrumpkinPoint,
+};
+
+/// Re-exports that give access to the standard aliases used in the code base, for bn256
+pub mod bn256 {
```
leonardoalt, line 31: @huitseeker maybe this file should be renamed to `bn256_grumpkin` to be consistent with halo2curves

huitseeker: The types that this is exporting are all defined there: https://github.com/privacy-scaling-explorations/halo2curves/blob/main/src/bn256/mod.rs
Do you mean the module at the top of the file or the line you're commenting on?

PR #184, opened 2023-06-19
[refactorings] Leftovers (pot-pourri?)

huitseeker:
I didn't quite have a place to put those as I was reading / refactoring other bits: opinions on those loosely held, happy to adjust.

The commit are tailored to be independent, and the individual commit messages are meant to be informative.

No serious effect on benchmarks, one way or another:
https://gist.github.com/huitseeker/93c66fda2f8444870e9374da990de541


srinathsetty (approved):
Looks great! thank you @huitseeker!

Issue #146, opened 2023-03-06
Add Supernova Support


wyattbenno777:
I would like to help add Supernova if not done already. Are you guys planning on adding it?? @srinathsetty

huitseeker:
This looks like a duplicate of #120 .

wyattbenno777:
People keep asking for this 👇 

I started an implmentation in a fork.
There is a lot of code simply reused and tweaked from Nova.
R1CS is first but I am following closely on the PSE CCS implementation.

https://github.com/wyattbenno777/SuperNova/tree/supernova

Let me know if someone is also doing this! 🙏

- [x] Modelling Runnng Instances
- [x] Added U_i, pci and hash(pci, Ui, zi) as public outputs.
- [ ] 4.4 Optimizations from the paper for representing U_i
- [ ] Clean up and rebase with the main Nova branch

srinathsetty:
Closing this issue since there is a pending PR on this topic.

PR #204, opened 2023-07-16
supernova implementation with naive opcode commitment for argumented circuit sequence.

hero78119:
Related to issue https://github.com/microsoft/Nova/issues/146

This works majorly based on the [initial works](https://github.com/microsoft/Nova/issues/146#issuecomment-1631883880) credits to @wyattbenno777, who initiatives a neat and clean skeleton then make further refactor possible! 🥇 

> To @wyattbenno777: there are many subtleties related to initial works on [folk](https://github.com/wyattbenno777/SuperNova/tree/supernova) and motivate me to refactor it accordingly :) Your commit is cherry-pick and included/preserved in this PR. (Some of commit is modified due to touch CCS on PSE Nova repo, therefore I clean up to isolated supernova features). Feel free to comments/raise question and we can discuss here with more clear manner!

This implementation idea is mostly followed supernova paper https://eprint.iacr.org/2022/1758, with some modified details described as below.
Feature flagged under a `supernova` flag that is off by default, so it has no impact on other parts of the code or users of the Nova library.
Comments are welcome!

### consolidate `Φ(zi, wi)` functionality into step_circuit/F circuit
In realistic ISA design, program counter can be set as arbitrary value by opcode itself. 
Take [RIV32 JAL](https://jemu.oscc.cc/JAL) opcode for example, `pc_{i}` will be added by `imm` offset to obtain `pc_{i+1}`.
In other words, `pc_{i+1}` manipulation makemore sense to be constrained by correspondence opcode implementation, i.e. step_circuit. 
This PR make above feature possible, by enhancing `synthesis` function to pass `pc_i` along with `z_i` as input, and output corresponding `pc_{i+1}` and `z_{i+1}` in supernova, and `pc_{i}` will be assumed to be checked by step circuit, along with a memory commitment to prove memory[pc_{i}] matched with current opcode. A Read-only-Memory, ROM is introduced as a naive memory commitment demo. More detail of ROM will be elaboration below. 

### extract circuit index selector against program counter, execution sequence correctness
In supernova paper, one of implicit part is how to choose i+1 argumented F' circuit to fold previous relaxed r1cs and r1cs.  
To assure next F' argumented circuit is invoked in pre-defined sequence, each F' circuit should constraint it order with each pc_i. 

This implementation demo a naive prototype of RAM machine: a Read-Only-Memory, ROM for opcode sequence constraints. ROM is passed as public input z as a flattern vector, [<stepcircuit_1>, <stepcircuit_2>, …], then each step_circuit constraint `rom[pc_{i}] == step_circuit_index`. In the later works, it should be replaced by more advanced memory commitment, such as merkle root + merkle proofs, or other batched [KZG commitment](https://dankradfeist.de/ethereum/2020/06/16/kate-polynomial-commitments.html). This can be combined together as `4.4 Optimizations, p27` in supernova paper. 

### secondary circuit logic
For supernova secondary circuit, program counter just assign 0 for each folding step. There is no non-trivial logic in supernova secondary circuit.
 
### Enhance debugging ability
This PR also introduces debugging friendly feature to `is_sat` and `is_sat_relaxed` function to point out failed constrain’s correspondence annotation. This improve a lot comparing with show true/false only.

## Follow-up works (thought can be future PRs)
- more documentation on supernova
- introduce vector commitment, address RAM machine memory commitment and `4.4 Optimizations, p27` in supernova paper.


hero78119:

srinathsetty:
Thanks for the PR! Given this is a relatively big code contribution, it will take some time to review it.

hero78119:
> Thanks for the PR! Given this is a relatively big code contribution, it will take some time to review it.

Well noted and thanks for update! 
Besides, added 2 more commits to optimise recursive-snark proof size from 2*running_r1cs to running_r1cs + 1. Primary circuit can just fold single running r1cs for each iteration, only secondary circuit need to fold respective strictly r1cs by argumented circuit index :)

src/supernova/mod.rs
```diff
@@ -0,0 +1,1232 @@
...
+  F_arity_primary: usize,
+  F_arity_secondary: usize,
+  ro_consts_primary: ROConstants<G1>,
+  ro_consts_circuit_primary: ROConstantsCircuit<G2>,
+  ck_primary: Option<CommitmentKey<G1>>,
+  r1cs_shape_primary: R1CSShape<G1>,
+  ro_consts_secondary: ROConstants<G2>,
+  ro_consts_circuit_secondary: ROConstantsCircuit<G1>,
+  ck_secondary: Option<CommitmentKey<G2>>,
+  r1cs_shape_secondary: R1CSShape<G2>,
```
huitseeker, line 86: This seems to be precisely the same struct at `PublicParams`. Would some sizing of the (R1CS) contents of the ROM be more appropriate than `r1cs_shape_primary` and `r1cs_shape_secondary` here?

src/supernova/mod.rs
```diff
@@ -0,0 +1,1232 @@
...
+      None,
+      c_primary,
+      ro_consts_circuit_primary.clone(),
+      u_i_length,
+    );
+    let mut cs: ShapeCS<G1> = ShapeCS::new();
+    let _ = circuit_primary.synthesize(&mut cs);
+
+    // We use the largest commitment_key for all instances
+    let r1cs_shape_primary = cs.r1cs_shape_without_commitmentkey();
```
huitseeker, line 131: How is the claim of `largest` supported?

huitseeker (commented):
I haven't looked at the circuit logic in detail yet and am for now just checkpointing in-line suggestions which I think would be helpful. heterogeneity is super exciting, thanks for opening a PR in that direction!

src/supernova/mod.rs
```diff
@@ -0,0 +1,1232 @@
...
+
+mod circuit; // declare the module first
+use circuit::{CircuitInputs, CircuitParams, SuperNovaCircuit};
+
+fn compute_digest<G: Group, T: Serialize>(o: &T) -> G::Scalar {
+  // obtain a vector of bytes representing public parameters
+  let mut encoder = ZlibEncoder::new(Vec::new(), Compression::default());
+  bincode::serialize_into(&mut encoder, o).unwrap();
+  let pp_bytes = encoder.finish().unwrap();
+
```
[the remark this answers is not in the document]
hero78119: Have ported in latest commit

src/gadgets/utils.rs
```diff
@@ -432,3 +432,22 @@ pub fn select_num_or_one<F: PrimeField, CS: ConstraintSystem<F>>(
...
+) -> Result<AllocatedNum<F>, SynthesisError> {
+  let c = AllocatedNum::alloc(cs.namespace(|| "c"), || {
+    Ok(*a.get_value().get()? + b.get_value().get()?)
+  })?;
+  cs.enforce(
+    || "Check u_fold",
+    |lc| lc,
+    |lc| lc,
+    |lc| lc + c.get_variable() - a.get_variable() - b.get_variable(),
+  );
```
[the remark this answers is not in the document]
hero78119: agree, thanks for point out that
Fixed in latest commit

src/r1cs.rs
```diff
@@ -182,18 +185,29 @@ impl<G: Group> R1CSShape<G> {
...
+            Ok(())
+          }
+        })
+        .collect::<Result<Vec<()>, _>>();
 
-      res == 0
+      match res {
+        Ok(_) => Ok(()),
+        Err(e) => Err(e),
+      }
```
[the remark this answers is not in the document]
hero78119: Fixed and adopt this syntax in latest commit

src/r1cs.rs
```diff
@@ -203,11 +217,14 @@ impl<G: Group> R1CSShape<G> {
...
 
-    if res_eq && res_comm {
-      Ok(())
-    } else {
-      Err(NovaError::UnSat)
+    if res_eq.is_err() {
+      return res_eq;
+    }
+    if !res_comm {
+      println!("res_comm not true");
```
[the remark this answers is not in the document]
hero78119: y and clean up

src/r1cs.rs
```diff
@@ -221,28 +238,39 @@ impl<G: Group> R1CSShape<G> {
...
+            // error
+            Err(self._get_constraints_path_by_index(i))
+          } else {
+            Ok(())
+          }
+        })
+        .collect::<Result<Vec<()>, _>>();
 
-      res == 0
+      res.map(|_| ()) // cleanup as Ok(())
```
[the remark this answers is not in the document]
hero78119: Fixed

src/gadgets/r1cs.rs
```diff
@@ -333,45 +333,50 @@ impl<G: Group> AllocatedRelaxedR1CSInstance<G> {
...
+  a: &AllocatedRelaxedR1CSInstance<G>,
+  b: &AllocatedRelaxedR1CSInstance<G>,
+  condition: &Boolean,
+) -> Result<AllocatedRelaxedR1CSInstance<G>, SynthesisError> {
+  let c = AllocatedRelaxedR1CSInstance {
+    u: conditionally_select(cs.namespace(|| "selected u"), &a.u, &b.u, condition)?,
+    W: conditionally_select_point(cs.namespace(|| "selected W"), &a.W, &b.W, condition)?,
+    E: conditionally_select_point(cs.namespace(|| "selected E"), &a.E, &b.E, condition)?,
+    X0: conditionally_select_bignat(cs.namespace(|| "selected X0"), &a.X0, &b.X0, condition)?,
+    X1: conditionally_select_bignat(cs.namespace(|| "selected X1"), &a.X1, &b.X1, condition)?,
```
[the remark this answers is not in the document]
hero78119: Yes, revamped all of them to `<out> = cond ? <a> : <b>` format

src/supernova/circuit.rs
```diff
@@ -0,0 +1,767 @@
...
+  },
+  Circuit, ConstraintSystem, SynthesisError,
+};
+use ff::Field;
+use serde::{Deserialize, Serialize};
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct CircuitParams {
+  limb_width: usize,
+  pub n_limbs: usize,
```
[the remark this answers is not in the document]
hero78119: This `n_limbs` will be used in `RecursiveSNARK.verify` ROCircuit to calculate number of field to RO, instead of hardcod. 
Fixed it to private variable, with a getter `get_n_limbs`

src/supernova/circuit.rs
```diff
@@ -0,0 +1,767 @@
...
+//!    2. Instance is folded into Ui[pci] correctly; just like Nova.
+//!    3. Invokes the function ϕ on input (zi, ωi) to compute pci+1, which represents
+//!    the index of the function Fj currently being run. pci+1 is then sent to the
+//!    next invocation of an augmented circuit (which contains a verifier circuit).
+
+//! There are two Verification Circuits. The primary and the secondary.
+//! Each of them is over a Pasta curve but
+//! only the primary executes the next step of the computation.
+//! We have two running instances. Each circuit takes as input 2 hashes: one for each
+//! of the running instances. Each of these hashes is
```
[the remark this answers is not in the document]
hero78119: Fixed, all the supernova docs/comment are preliminary rephrased in latest commit.

src/r1cs.rs
```diff
@@ -221,28 +238,39 @@ impl<G: Group> R1CSShape<G> {
...
-    } else {
-      Err(NovaError::UnSat)
+    if res_eq.is_err() {
+      return res_eq;
+    }
+    if !res_comm {
+      println!("res_comm not true");
+      return Err(NovaError::UnSat);
     }
+    Ok(())
```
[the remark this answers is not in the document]
hero78119: clean up in latest commit

src/bellperson/r1cs.rs
```diff
@@ -53,9 +55,17 @@ where
   G::Scalar: PrimeField,
 {
   fn r1cs_shape(&self) -> (R1CSShape<G>, CommitmentKey<G>) {
+    let S = self.r1cs_shape_without_commitmentkey();
+    let ck = R1CS::<G>::commitment_key(&S);
+
+    (S, ck)
+  }
```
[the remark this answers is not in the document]
hero78119: Nice suggestion.
Besides, also rename functions under `NovaShape` trait
- `r1cs_shape` -> `r1cs_shape_with_commitmentkey`
- `r1cs_shape_without_commitmentkey` -> `r1cs_shape`

Hope it make more self-explanation

src/r1cs.rs
```diff
@@ -31,6 +31,7 @@ pub struct R1CSShape<G: Group> {
   pub(crate) num_cons: usize,
   pub(crate) num_vars: usize,
   pub(crate) num_io: usize,
+  pub(crate) constraints_path: Option<Vec<String>>,
```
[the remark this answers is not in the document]
hero78119: I opt this part of memory usage, by 
1. `is_sat/is_sat_relaxed` will return just `index: usize` of constraint error
2. then the `usize -> String` will be happened on verify() function.
3. There is no `ShapeCS` reference holding and carry over, instead the `constraint_path: Vec<String>` is put in `PublicParams` and setup only once
4. remove  `derived(clone)` from PublicParams, to avoid abuse

Which  I think is the better handling for now, with minimal interface change to achieve the happy debugging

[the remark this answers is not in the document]
hero78119: Beside I also fine tune running instances vector copy by `to_vec()` function. I think now memory opt a lot

src/r1cs.rs
```diff
@@ -203,11 +217,14 @@ impl<G: Group> R1CSShape<G> {
       U.comm_W == comm_W && U.comm_E == comm_E
     };
 
-    if res_eq && res_comm {
-      Ok(())
-    } else {
-      Err(NovaError::UnSat)
+    if res_eq.is_err() {
+      return res_eq;
+    }
```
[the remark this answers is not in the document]
hero78119: fixed and tks for suggestion

src/r1cs.rs
```diff
@@ -346,11 +375,27 @@ impl<G: Group> R1CSShape<G> {
...
+
+  fn _get_constraints_path_by_index(&self, index: usize) -> NovaError {
+    // error
+    if let Some(constraints_path) = &self.constraints_path {
+      NovaError::UnSatMsg(format!(
+        "is_relaxed_sat relation failed at constraint path {:?}",
+        constraints_path
+          .get(index)
+          .clone()
+          .unwrap_or(&"".to_string())
```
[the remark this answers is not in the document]
hero78119: fixed

src/r1cs.rs
```diff
@@ -221,28 +238,39 @@ impl<G: Group> R1CSShape<G> {
...
     // verify if comm_W is a commitment to W
     let res_comm: bool = U.comm_W == CE::<G>::commit(ck, &W.W);
 
-    if res_eq && res_comm {
-      Ok(())
-    } else {
-      Err(NovaError::UnSat)
+    if res_eq.is_err() {
+      return res_eq;
+    }
```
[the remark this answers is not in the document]
hero78119: fixed

src/supernova/mod.rs
hero78119: `r1cs_shape_primary` and `r1cs_shape_secondary` are not the same struct, since secondary circuit is just a trivial circuit and do nothing while primary circuit got the application logic. Not sure I understand correctly, would you kindly elaborate more on this :)

src/supernova/circuit.rs
```diff
@@ -0,0 +1,767 @@
...
+    secondary_params: CircuitParams,
+    ro_consts1: ROConstantsCircuit<G2>,
+    ro_consts2: ROConstantsCircuit<G1>,
+    num_constraints_primary: usize,
+    num_constraints_secondary: usize,
+  ) where
+    G1: Group<Base = <G2 as Group>::Scalar>,
+    G2: Group<Base = <G1 as Group>::Scalar>,
+  {
+    /*
```
srinathsetty, line 651: Nit: It might be preferrable to use // instead of /* ... */

src/supernova/mod.rs
```diff
@@ -0,0 +1,1194 @@
...
+    z0_primary: &Vec<G1::Scalar>,
+    z0_secondary: &Vec<G2::Scalar>,
+    ck_primary: &CommitmentKey<G1>,
+    ck_secondary: &CommitmentKey<G2>,
+  ) -> Result<Self, NovaError> {
+    if z0_primary.len() != pp.F_arity_primary || z0_secondary.len() != pp.F_arity_secondary {
+      return Err(NovaError::InvalidInitialInputLength);
+    }
+
+    match recursive_snark {
```
srinathsetty, line 253: This code might be based on an older version of src/lib.rs? I think there was some non-trivial improvement from using mutable types. Is it worth porting that code here?

huitseeker: FYI, this is about #163

CPerezz:
Hmmm.

It's hard for me to see how this fits in inside of this lib instead of being a standalone crate. 
We already have a bunch of code to maintain like the spartan re-implementation with SPARK that is "duplicated" from the original Spartan code.

SuperNova is actually a nice use-case to have `nova-snark` as a dependency. Which will help to also see API visibility concerns.
Not sure is the best choice to merge this here rather than have it as a standalone crate that depends on this.

src/supernova/mod.rs
hero78119: Thanks for the hints. Fixed in latest commit and align with Nova performance fix PR https://github.com/microsoft/Nova/pull/163

src/supernova/circuit.rs
hero78119: Fixed

hero78119:
> Hmmm.
> 
> It's hard for me to see how this fits in inside of this lib instead of being a standalone crate. We already have a bunch of code to maintain like the spartan re-implementation with SPARK that is "duplicated" from the original Spartan code.
> 
> SuperNova is actually a nice use-case to have `nova-snark` as a dependency. Which will help to also see API visibility concerns. Not sure is the best choice to merge this here rather than have it as a standalone crate that depends on this.

From my viewpoint I think supernova is a generalized version of current Nova implementation (Nova is a special case for 1 augmented circuit version)
Feel like that, if SuperNova benchmark also show no much different comparing with Nova, Maybe it make sense to have Supernova as a standalone version

src/supernova/mod.rs
```diff
@@ -0,0 +1,1142 @@
...
+        .unwrap_or_else(|| initial_program_counter);
+      let augmented_circuit_index = rom[u32::from_le_bytes(
+        // convert program counter from field to usize (only took le 4 bytes)
+        program_counter.to_repr().as_ref().clone()[0..4]
+          .try_into()
+          .unwrap(),
+      ) as usize];
+
+      let mut recursive_snark = recursive_snark_option.unwrap_or_else(|| {
+        if augmented_circuit_index == OPCODE_0 {
```
hero78119, line 1068: This demo have some code duplication on opcode multiplexing, due to slightly hard on handling heterogeneous type of augment circuit. I try few options to beautify this part and no satisfied version. 

If you have better suggestion kindly let me know, thanks

wyattbenno777:
> Hmmm.
> 
> It's hard for me to see how this fits in inside of this lib instead of being a standalone crate. We already have a bunch of code to maintain like the spartan re-implementation with SPARK that is "duplicated" from the original Spartan code.
> 
> SuperNova is actually a nice use-case to have `nova-snark` as a dependency. Which will help to also see API visibility concerns. Not sure is the best choice to merge this here rather than have it as a standalone crate that depends on this.

My opinion is that it would belong here as a 'feature'. 
For many use-case the a'la carte benefits of SuperNova are needed. 

Other things I would include:
1. Different cycles of curves. Like @huitseeker is already working on.
(in the future possibility to remove cycles for another method .)

2. Different/future final SNARK wrappers.

A user would come and say: "I want a Nova proof of xyz, the verifier needs to work on xyz blockchain/environment",
and be able to configure that fairly easily. 

My 2 cents.

srinathsetty:
> > Hmmm.
> > It's hard for me to see how this fits in inside of this lib instead of being a standalone crate. We already have a bunch of code to maintain like the spartan re-implementation with SPARK that is "duplicated" from the original Spartan code.
> > SuperNova is actually a nice use-case to have `nova-snark` as a dependency. Which will help to also see API visibility concerns. Not sure is the best choice to merge this here rather than have it as a standalone crate that depends on this.
> 
> From my viewpoint, supernova is a generalized version of current Nova implementation (Nova is a special case for 1 augmented circuit version) Feel like that, if SuperNova benchmark also show no much different comparing with Nova, Maybe it make sense to have Supernova as a standalone version, in separate PR maybe.
> 
> Would like to hear more opinions from author later :)

@CPerezz, regarding a stand-alone crate with nova-snark as a dependency, wouldn't it need access to several internal methods of Nova (e.g., nifs.rs), which is not currently exposed? This would also require nova to move to being a workspace right? Or, is there a different and better approach to multiple packages?

In general, I do like the layering approach, but I will have to look through the new code to see what code will need to be exposed for the layering to work.

hero78119:
Updated: 
Optimize supernova RecursiveSNARK prove/verify step to pass by reference almost everywhere, around 10% improvement based on benchmark.

Also run [benchmark recursive-snark](https://gist.github.com/hero78119/5e955ddd51135b12a7bf26f54148dd17) for Nova vs Supernova-1-augmented-circuit version  
In summary, in low constraints case no much different, on high constraints case Supernova is around 5-10% slower. However  on largest constraints `1038757` supernova is 10% fast. 

Benchmark is on windows-subsystem-linux so perf will be variant. Will be better to run benchmark on other neutral server to get more accurate result.

src/supernova/circuit.rs
```diff
@@ -0,0 +1,754 @@
...
+}
+
+#[derive(Debug)]
+pub struct CircuitInputs<'a, G: Group> {
+  params_next: G::Scalar,
+  params: G::Scalar,
+  i: G::Base,
+  z0: &'a Vec<G::Base>,
+  zi: Option<&'a Vec<G::Base>>,
+  U: Option<&'a Vec<Option<RelaxedR1CSInstance<G>>>>,
```
huitseeker, line 73: Those references of vectors are the hallmark of a place where you'd want to use slices (`&[G::Base]`). Running clippy should help you find those opportunities.

huitseeker: FYI, the lint you're looking to activate is [`ptr_arg`](https://rust-lang.github.io/rust-clippy/master/#/ptr_arg)

src/supernova/mod.rs
```diff
@@ -0,0 +1,1142 @@
...
+  use bellperson::gadgets::boolean::Boolean;
+  use core::marker::PhantomData;
+  use ff::Field;
+  use ff::PrimeField;
+
+  use super::*;
+
+  use ::bellperson::{gadgets::num::AllocatedNum, ConstraintSystem, SynthesisError};
+
+  fn constraint_curcuit_index<F: PrimeField, CS: ConstraintSystem<F>>(
```
huitseeker, line 757: Typo.

src/supernova/circuit.rs
hero78119: Thanks for the hint, applied clippy and fix all warning/errors

hero78119:
Updated: fixed a soundness in supernova implementation. 

More detail: in Nova, `digest` value of publicparam is constraint to be the same for each folding round, since there is uniform IVC
.
In Supernova Non-Uniform IVC, there will be > 1 augmented circuit, each will have different primary circuit R1CS shape, and in each round of folding, different augmented circuit will be invoked and fold to get {i+1} running instance. Soundness is introduced in this implementation as there allow a `digest_{i}/digest_{i+1}` be witness, which allows to fill different digest value under SAME augmented index in the intermediated folding rounds. 

In latest commit, this is fixed by have a `unified_digest = compute_digest([pp1, pp2, ...])` and removing digest for each augmented circuit.

hero78119:
Update: rebase to latest branch and code cosmetics 

> Updated: Optimize supernova RecursiveSNARK prove/verify step, circuit input,... to pass by reference almost everywhere, around 10% improvement based on benchmark.
> 
> Also run [benchmark recursive-snark](https://gist.github.com/hero78119/5e955ddd51135b12a7bf26f54148dd17) for Nova vs Supernova-1-augmented-circuit version In summary, in low constraints case no much different, on high constraints case Supernova is around 5-10% slower. However on largest constraints `1038757` supernova is 10% fast.
> 
> Benchmark is on windows-subsystem-linux so perf will be variant. Will be better to run benchmark on other neutral server to get more accurate result.

More update: Compare with Nova, Supernova implementation have extra circuit logic on `NIFS.V` to pick correct running instances and fold with. incomming r1cs instance on secondary circuit. The circuit size will grow linear  O(#num_of_argumented_circuit). Therefore, benchmark shows that in single augmented circuit case, Supernova is slower around 10%

Giving two design choices
1. Nova + one augment circuit, and have one uniform step circuit to handle non-uniform logic multiplexing inside.
2. Supernova + multiple augment circuit, each step circuit is as simple as possible to do one thing, and multiplexing handle by Supernova folding.

A scenario for choose 2 over 1 is, if your application's operations is distributed non-uniformly, such that some complex operations will be invoked infrequent, and more proportion is on simple operations. This case Supernova outperforms, because you paid a slightly overhead on `NIFS.V` circuit, however benefit a lot as most operation are invoke on smaller shape step circuit with less constraints, thus less MSM.  This is especially useful on `stateful machine with a particular instruction set`

src/supernova/mod.rs
huitseeker: Setting heterogeneity aside for a moment, the sizing of the public parameters depends on the shape of the circuit they are used for. In particular, you would want the `CommitmentKey` (in practice, a vector Pedersen) to offer enough generators for the size of the step circuit. How this is done is expanded upon in PR #203 .

With heterogeneity, I would expect that same dependency to still exist, but the circuit bookkeeping to be different. More specifically, I would expect the `R1CSShape` of the *largest* circuit among those in the ROM to drive the size requirements for the public parameters.

huitseeker: This was hinted at in a [prior comment](https://github.com/microsoft/Nova/pull/204#discussion_r1267328957). 
Nova, as a library, is quite safe in sizing public parameters:
- Nova runs through circuit synthesis (which is costly!) to size the public parameters precisely, see [here](https://github.com/microsoft/Nova/blob/f0b27b7457556f58cffb01e6c89b5c7e159a694f/src/lib.rs#L110-L121).
- see #203 for other nitty-gritty details.

The approach of simply recommending the user to pass the largest circuit among those in the ROM here may be error-prone -- as I'm worried a user might miss the requirement.

src/supernova/mod.rs
```diff
@@ -0,0 +1,1202 @@
...
+      })
+      .collect::<Vec<Option<RelaxedR1CSWitness<G1>>>>();
+
+    let r_U_primary_initial_list = (0..num_augmented_circuits)
+      .map(|i| {
+        if i == first_augmented_circuit_index {
+          Some(r_U_primary.clone())
+        } else {
+          None
+        }
```
huitseeker, line 390: Nit: `(i == first_augmented_circuit_index).then(|| r_U_primary.clone())`

src/supernova/mod.rs
```diff
@@ -0,0 +1,1202 @@
...
+      .get_value()
+      .ok_or(NovaError::SynthesisError)?;
+    let zi_secondary = zi_secondary
+      .iter()
+      .map(|v| v.get_value().ok_or(NovaError::SynthesisError))
+      .collect::<Result<Vec<<G2 as Group>::Scalar>, NovaError>>()?;
+
+    if zi_primary.len() != pp.F_arity_primary || zi_secondary.len() != pp.F_arity_secondary {
+      return Err(NovaError::InvalidStepOutputLength);
+    }
```
huitseeker, line 569: Same idea: this check is probably better placed as early as we can.

src/supernova/mod.rs
```diff
@@ -0,0 +1,1202 @@
...
+        "hash_primary {:?} not equal l_u_secondary.X[0] {:?}",
+        hash_primary, self.l_u_secondary.X[0]
+      );
+      return Err(NovaError::ProofVerifyError);
+    }
+    if hash_secondary != scalar_as_base::<G2>(self.l_u_secondary.X[1]) {
+      println!(
+        "hash_secondary {:?} not equal l_u_secondary.X[1] {:?}",
+        hash_secondary, self.l_u_secondary.X[1]
+      );
```
huitseeker, line 694: The `println!` statements are very understandable (I've inserted similar ones in that position in my local branches myself), but we probably don't want them here. Consider using the `log` crate, with a `debug` log-level instead.

src/supernova/mod.rs
```diff
@@ -0,0 +1,1202 @@
...
+  /// verify recursive snark
+  pub fn verify<C1: StepCircuit<G1::Scalar>, C2: StepCircuit<G2::Scalar>>(
+    &mut self,
+    claim: &RunningClaim<G1, G2, C1, C2>,
+    z0_primary: &[G1::Scalar],
+    z0_secondary: &[G2::Scalar],
+  ) -> Result<(), NovaError> {
+    // number of steps cannot be zero
+    if self.i == 0 {
+      println!("must verify on valid RecursiveSNARK where i > 0");
```
huitseeker, line 595: Consider `log` for this sort of statement (at the error level, here).

src/supernova/mod.rs
```diff
@@ -0,0 +1,1202 @@
...
+          println!("r_U_primary got instance length {:?} != {:?}", U.X.len(), 2);
+          Err(NovaError::ProofVerifyError)
+        } else {
+          Ok(())
+        }
+      })?;
+
+    self.r_U_secondary[0].as_ref().map_or(Ok(()), |U| {
+      if U.X.len() != 2 {
+        println!(
```
huitseeker, line 624: Please use `log::debug!` here instead.

src/supernova/mod.rs
```diff
@@ -0,0 +1,1202 @@
...
+        assert!(res.is_ok());
+      } else if augmented_circuit_index == OPCODE_1 {
+        let res = recursive_snark.prove_step(&running_claim2, &z0_primary, &z0_secondary);
+        if let Err(e) = &res {
+          println!("res failed {:?}", e);
+        }
+        assert!(res.is_ok());
+        let res = recursive_snark.verify(&running_claim2, &z0_primary, &z0_secondary);
+        if let Err(e) = &res {
+          println!("res failed {:?}", e);
```
huitseeker, line 1172: In a unit test, consider `.unwrap()` for errors rather than printing their contents and then asserting on ok.

src/supernova/mod.rs
```diff
@@ -0,0 +1,1202 @@
...
+    })
+  }
+  /// executing a step of the incremental computation
+  pub fn prove_step<C1: StepCircuit<G1::Scalar>, C2: StepCircuit<G2::Scalar>>(
+    &mut self,
+    claim: &RunningClaim<G1, G2, C1, C2>,
+    z0_primary: &[G1::Scalar],
+    z0_secondary: &[G2::Scalar],
+  ) -> Result<(), NovaError> {
+    // Frist step was already done in the constructor
```
huitseeker, line 417: Nit: typo

src/supernova/mod.rs
```diff
@@ -0,0 +1,1202 @@
...
+          ck_primary,
+          &pp.ro_consts_primary,
+          &self.pp_digest,
+          &pp.r1cs_shape_primary,
+          &RelaxedR1CSInstance::default(ck_primary, &pp.r1cs_shape_primary),
+          &RelaxedR1CSWitness::default(&pp.r1cs_shape_primary),
+          &l_u_primary,
+          &l_w_primary,
+        )?
+      };
```
huitseeker, line 522: Nit: This might look slightly better as
```rust
match (self.r_U_primary.get(claim.get_augmented_circuit_index()), self.r_W_primary.get(claim.get_augmented_circuit_index())) {
  Some((r_U_primary, r_W_primary)) => ...,
  _ => ...
}
```

Just a thought (what you already have is fine)

Cargo.toml
```diff
@@ -66,10 +66,15 @@ harness = false
 name = "sha256"
 harness = false
 
+[[bench]]
+name = "recursive-snark-supernova"
+harness = false
+
```
huitseeker, line 72: I think you want `required-features = ["supernova"]` here

huitseeker (commented):
I'm excited about this PR, it's really nice to see this laid out - and I hope my inline comments may help. When I say "nit", it means my comment is stylistic nitpicking that you can safely ignore. 

I like the demonstration of the setup in `test_trivial_nivc_with`, specifically the sizing of the commitment key there. I would love to see this supported a bit better by taking the ad-hoc ROM struct and its loosely coupled `test_circuit{1,2}` defined in this test, and deducing a public struct and API that would let an end-user perform the same work in a more "managed" way.

With that said, the optional "supernova" feature may offer a way to figure this out in a future PR.

porcuquine (commented):
Thanks for working on this. I haven't had enough time to fully absorb this, but after a first partial pass, I did have some initial questions and thoughts, which I'll submit now while fresh.

src/supernova/mod.rs
```diff
@@ -0,0 +1,1202 @@
...
+            &agg,
+          )
+        })?;
+
+    cs.enforce(
+      || "selected_circuit_index == circuit_index",
+      |lc| lc + selected_circuit_index.get_variable(),
+      |lc| lc + CS::one(),
+      |lc| lc + circuit_index.get_variable(),
+    );
```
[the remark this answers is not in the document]
hero78119: Hi, firstly thanks for review and nice catch for soundness! Yes I aware that when start integrating Supernova into other project. Thanks again for point out this and reminding for the fix :)

I think the key point here is we need to define and encode `constant` into the circuit. In current R1CS design the only const is the F::ONE, so one way to derived other constants safely is compiled them the circuit from F::ONE via bit decomposition, double then sum.

PR is updated in commit https://github.com/microsoft/Nova/pull/204/commits/ebe1f08a84c132324a09a55e3add3e8903c462fe . To summary the fixs

- `circuit_index` should be a circuit const
- `indexes_alloc` should also be incremental circuit const. (It was also  implemented efficiently by add one by one from F::ONE)
- then all the constraints should derived from above circuit const

Hope the new fix the problem. please kindly review, and thanks again for pointing out this. Really like the your mindset to check the potiential under-constraint.  

I just replied in this comment for other comments above all points out the same issue :)

src/gadgets/utils.rs
```diff
@@ -74,6 +77,103 @@ pub fn alloc_one<F: PrimeField, CS: ConstraintSystem<F>>(
...
+      Ok(value)
+    })?;
+  let sum_lc = LinearCombination::zero() + value_alloc.get_variable() - &sum;
+  cs.enforce(
+    || format!("{:?} sum - value = 0", value),
+    |lc| lc + &sum_lc,
+    |lc| lc + CS::one(),
+    |lc| lc,
+  );
+  Ok(value_alloc)
```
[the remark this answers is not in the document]
porcuquine: It's true that what you've done here enforce a maximum number of bits — but at the cost of that number of constraints, so if the goal is only constraint efficiency it seems unnecessary. Is there another reason to introduce the bit limit?

porcuquine (commented):
This is looking better. I had a few more questions and comments about implementation details, but the big issue seems to have been fixed.

src/supernova/mod.rs
```diff
@@ -0,0 +1,1202 @@
...
+      .get_value()
+      .ok_or(NovaError::SynthesisError)?;
+    let zi_secondary = zi_secondary
+      .iter()
+      .map(|v| v.get_value().ok_or(NovaError::SynthesisError))
+      .collect::<Result<Vec<<G2 as Group>::Scalar>, NovaError>>()?;
+
+    if zi_primary.len() != pp.F_arity_primary || zi_secondary.len() != pp.F_arity_secondary {
+      return Err(NovaError::InvalidStepOutputLength);
+    }
```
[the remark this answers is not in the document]
hero78119: good idea, updated in https://github.com/microsoft/Nova/pull/204/commits/61dad5b531b7316313b991d9651a32584286a6c6 (I think nova side also need to apply the same suggestion, can be separate PR)

src/supernova/mod.rs
hero78119: ditto and fixed in https://github.com/microsoft/Nova/pull/204/commits/61dad5b531b7316313b991d9651a32584286a6c6

src/supernova/mod.rs
hero78119: Tried to switch to log crate in commit https://github.com/microsoft/Nova/pull/204/commits/61dad5b531b7316313b991d9651a32584286a6c6

src/supernova/mod.rs
hero78119: done with commit https://github.com/microsoft/Nova/pull/204/commits/61dad5b531b7316313b991d9651a32584286a6c6

src/supernova/mod.rs
hero78119: Adopted the way you suggest which makes code shorter and clean 👍

src/gadgets/utils.rs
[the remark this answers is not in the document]
hero78119: I over-complex stuff and after hints from Lurks `allocate_constant` I realized it just need to be constraints via coeeficient haha 😄 
Updated in commit https://github.com/microsoft/Nova/pull/204/commits/c9fb511c1dd3865fbf312879b5024ef6dacd3f29

src/supernova/mod.rs
```diff
@@ -790,46 +792,44 @@ mod tests {
...
-    let circuit_index = AllocatedNum::alloc(cs.namespace(|| "circuit_index"), || {
-      Ok(F::from(circuit_index as u64))
-    })?;
-    let pc_offset_in_z = AllocatedNum::alloc(cs.namespace(|| "pc_offset_in_z"), || {
-      Ok(pc_counter.get_value().unwrap() + F::from(rom_offset as u64))
-    })?;
+    let indexes_alloc = alloc_incremental_range_index(
+      cs.namespace(|| "augment circuit range index"),
+      0,
+      z[rom_offset..].len(),
```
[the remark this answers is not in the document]
hero78119: Nice suggestion. 
Update to pass `&z[rom_offset..]` from outside, also rename `z` to `rom` to express intent meaningfully.

srinathsetty:
Thanks @hero78119 for your prompt updates to the PR! I will review the PR more carefully and provide inputs as necessary.

src/supernova/mod.rs
```diff
@@ -0,0 +1,1178 @@
...
+        "hash_secondary {:?} not equal l_u_secondary.X[1] {:?}",
+        hash_secondary, self.l_u_secondary.X[1]
+      );
+      return Err(NovaError::ProofVerifyError);
+    }
+
+    // check the satisfiability of the provided `circuit_index` instance
+    let default_instance = RelaxedR1CSInstance::default(ck_primary, &pp.r1cs_shape_primary);
+    let default_witness = RelaxedR1CSWitness::default(&pp.r1cs_shape_primary);
+    let (res_r_primary, (res_r_secondary, res_l_secondary)) = rayon::join(
```
dajuguan, line 691: In Supernova's paper the verifier need to check every related instance-witness pair (U, W) satisfys  relaxed R1CS constrains. However, it seems only the claimed pair is checked here.

dajuguan (commented):
In Supernova's paper the verifier need to check every related instance-witness pair (U, W) satisfys  relaxed R1CS constrains. However, it seems only the claimed pair is checked here.

src/supernova/mod.rs
wyattbenno777: They are all verified in their own step. opcode 1 runs x times and is verified, then opcode 2. If the ROM is something like [1,2,1,1] it is run and verified for each step with the code you commented. At the very end a valid SuperNova proof would have both 1 and 2 be valid. 

https://github.com/microsoft/Nova/blob/ea14cabf116b6be34fee112737e00a64d8c81f7c/src/supernova/mod.rs#L1017C1-L1018C102

dajuguan: Thanks for your explanation.

src/supernova/mod.rs
hero78119: > This was hinted at in a [prior comment](https://github.com/microsoft/Nova/pull/204#discussion_r1267328957). Nova, as a library, is quite safe in sizing public parameters:
> 
> * Nova runs through circuit synthesis (which is costly!) to size the public parameters precisely, see [here](https://github.com/microsoft/Nova/blob/f0b27b7457556f58cffb01e6c89b5c7e159a694f/src/lib.rs#L110-L121).
> * see [feat: Make commitment key generation configurable with an optional parameter #203](https://github.com/microsoft/Nova/pull/203) for other nitty-gritty details.
> 
> The approach of simply recommending the user to pass the largest circuit among those in the ROM here may be error-prone -- as I'm worried a user might miss the requirement. If there was a ROM API in the form of an indexed iterator of circuits (perhaps with a cached way to determine their shape), I suspect it would make automating this part of the setup possible.

For current version, the automating way to `identify largest primary circuit` is relying on user to implement logic to sizing the largest circuit to setup commitment key based on it. demo in [Supernova test](https://github.com/hero78119/SuperNova/blob/supernova/src/supernova/mod.rs#L1056-L1073) 
I admitted the ideal way is to exploring way to hide it from user to be more error pruning.

src/supernova/mod.rs
```diff
@@ -0,0 +1,1202 @@
...
+#[serde(bound = "")]
+pub struct RecursiveSNARK<G1, G2>
+where
+  G1: Group<Base = <G2 as Group>::Scalar>,
+  G2: Group<Base = <G1 as Group>::Scalar>,
+{
+  r_W_primary: Vec<Option<RelaxedR1CSWitness<G1>>>,
+  r_U_primary: Vec<Option<RelaxedR1CSInstance<G1>>>,
+  r_W_secondary: Vec<Option<RelaxedR1CSWitness<G2>>>, // usually r_W_secondary.len() == 1
+  r_U_secondary: Vec<Option<RelaxedR1CSInstance<G2>>>, // usually r_U_secondary.len() == 1
```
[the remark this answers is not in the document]
hero78119: This is defined as vector for the sake of code sharing and no-copy in `prove` function.  
https://github.com/microsoft/Nova/blob/ea14cabf116b6be34fee112737e00a64d8c81f7c/src/supernova/mod.rs#L45

src/supernova/mod.rs
```diff
@@ -0,0 +1,1202 @@
...
+      }
+      if let Some(U) = self.r_U_secondary[0].as_ref() {
+        U.absorb_in_ro(&mut hasher)
+      }
+
+      let mut hasher2 =
+        <G1 as Group>::RO::new(pp.ro_consts_primary.clone(), num_field_secondary_ro);
+      hasher2.absorb(scalar_as_base::<G1>(self.pp_digest));
+      hasher2.absorb(G2::Scalar::from(self.i as u64));
+      hasher2.absorb(G2::Scalar::ZERO);
```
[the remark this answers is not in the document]
hero78119: It make sense to remove program counter from secondary circuit to save circuit size :)
removed in commit https://github.com/microsoft/Nova/pull/204/commits/2d3cce01982f55a5578d7a14e5a8750dda50288c

src/supernova/mod.rs
```diff
@@ -0,0 +1,1202 @@
...
+
+      for e in z0_primary {
+        hasher.absorb(*e);
+      }
+      for e in &self.zi_primary {
+        hasher.absorb(*e);
+      }
+      if let Some(U) = self.r_U_secondary[0].as_ref() {
+        U.absorb_in_ro(&mut hasher)
+      }
```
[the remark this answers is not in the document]
hero78119: Just update commit https://github.com/microsoft/Nova/pull/204/commits/2d3cce01982f55a5578d7a14e5a8750dda50288c to assure error prompt once None meet

src/supernova/mod.rs
srinathsetty: @wyattbenno777 I'm confused a bit by your explanation. In SuperNova instantiated over a cycle of curves, if there are \ell different step circuits, we will have \ell running instances (on the first curve), a running instance (on the second curve), and a fresh instance (second curve). We need to check that all \ell + 2 instances are satisfying. In the case of Nova, \ell = 1, so we have 3 instances verified in the main branch.

wyattbenno777:
In code, we are giving a PrimaryCircuit and a TrivialCircuit to a
RunningInstance object. It runs with these as many times as mentioned in
the ROM (there is an instance being folded in at each step). In the case of
Nova the user is still specifying a primary and a test circuit. So should I
be explaining this as 3 things to be verified for each opcode? (Primary,
Trivial, Thing to be folded on second curve) or as the things the user
specifies (Primary and Trivial circuit)?

src/supernova/mod.rs
wyattbenno777: In code, we are giving a PrimaryCircuit and a TrivialCircuit to a RunningInstance object. It runs with these as many times as mentioned in the ROM (there is an instance being folded in at each step). In the case of Nova the user is still specifying a primary and a trivial circuit. So should I be explaining this as 3 things to be verified for each opcode? (Primary, Trivial, Thing to be folded on second curve) or as the things the user specifies (Primary and Trivial circuit)?

wyattbenno777: Extra info for any future reader: pg. 15 https://eprint.iacr.org/2022/1758.pdf

hero78119: > They are all verified in their own step. opcode 1 runs x times and is verified, then opcode 2. If the ROM is something like [1,2,1,1] it is run and verified for each step with the code you commented. At the very end a valid SuperNova proof would have both 1 and 2 be valid.
> 
> https://github.com/microsoft/Nova/blob/ea14cabf116b6be34fee112737e00a64d8c81f7c/src/supernova/mod.rs#L1017C1-L1018C102

More addon. I think `opcode 1 runs x times` is indeed a bit confusing haha, because it sounds like we can run opcode 1 x times first then follow by opcode 2 runs y times, ....,  this mindset can explain well on Uniform-IVC Nova, but on Supernova, non-uniform with sequence need to be emphasised. 

Giving example of sequence [1, 2, 1, 1], Supernova assure 
- i=0, augmented circuit 1 is correctly invoked
- i=1, augmented circuit 2 is correctly invoked
- i=2, augmented circuit 1 ...
- i=3, augmented circuit 1 ...

and supernova need to capture soundness from the maliculous prover giving another sequence [1, 1, 1, 2]. Since order matters, and `y1 != y2, for y1=F_1(F_2(F_1(F_1(x)))), y2=F_1(F_1(F_1(F_2(x))))`.


Giving there are N opcodes on 2 cycles curve, there will be N running instances (on second curve) + 1 running instance (on first curve) + 1 fresh instance (second curve)

In each round of folding for the sake of efficiency, we can just verify **One** running instances (on second curve) + 1 running instance (on first curve) + 1 fresh instance (second curve).  The **One** is the opcode being folded in this round. Of course we can verify N running instance (on second curve) all together, but it's not nessesary since they will be verified on their own round. 

Actually, invoking `verify` on RecursiveSNARK (here verify is not the same as NIVC verify circuit!) at each round is more for demo purpose. In production I think we can just update RecursiveSNARK and have a final SNARK prove via CompressSNARK is enough (Terminlogies borrow from nova implementation, not paper)

wyattbenno777: This is a very well thought out answer!

*CycleNova will change this if/when implemented 😊

src/supernova/circuit.rs
```diff
@@ -0,0 +1,751 @@
...
+      T,
+      program_counter,
+      last_augmented_circuit_index,
+    }
+  }
+}
+
+/// The augmented circuit F' in SuperNova that includes a step circuit F
+/// and the circuit for the verifier in SuperNova's non-interactive folding scheme,
+/// SuperNova NIVS will fold strictly r1cs instance u with respective relaxed r1cs instance U[last_augmented_circuit_index]
```
porcuquine, line 109: NIVS -> NIFS?

src/supernova/circuit.rs
```diff
@@ -0,0 +1,751 @@
...
+    // The primary circuit just initialize single AllocatedRelaxedR1CSInstance
+    let U_default = if self.params.is_primary_circuit {
+      vec![AllocatedRelaxedR1CSInstance::default(
+        cs.namespace(|| "Allocate primary U_default".to_string()),
+        self.params.limb_width,
+        self.params.n_limbs,
+      )?]
+    } else {
+      // The secondary circuit convert the incoming R1CS instance on index which match last_augmented_circuit_index
+      let imcomming_r1cs = AllocatedRelaxedR1CSInstance::from_r1cs_instance(
```
porcuquine, line 268: 'incoming' not 'imcomming'

hero78119:
Update:
1. resolve conflicts with main branch
2. fix typo according to reviews
3. modify to fit ShapeCS/TestShapeCS usage

src/supernova/circuit.rs
```diff
@@ -0,0 +1,761 @@
...
+          U,
+          &empty_U,
+          &equal_bit,
+        )
+      })
+      .collect();
+
+    // Now reduce to one is safe, because only 1 of them != G::Zero based on the previous step
+    // TODO optimize this part to have raw linear-combination on variables to achieve less constraints
+    let U_to_fold = U?.iter().enumerate().try_fold(empty_U, |agg, (i, U)| {
```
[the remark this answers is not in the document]
hero78119: Hi, Thanks for catch up this and point out that!

Yes this is an under-constraints, due to there is no constraint for the `empty value` declaration.
I refered your idea and reviewed again of this logic.
Then I realized a better way to solve it is just rewrite this logic to avoid empty allocation, instead just iterate through the list to get the selected value when index match.

More detail in pseudo code
```
a := the array
index := target index
res = a[0] // default value is first element
res := index == 1 ? res:a[1]
res := index == 2 ? res:a[2] 
res := index == 3 ? res:a[3] 
...
return res
```
In particular, this ad-hoc N-reduce-1 optimisation is only applicable for this logic , giving assumption that prover have no choice but need to honestly choose correct witness `last_augmented_circuit_index` to hint the `fold(fresh instance, U[last_augmented_circuit_index])` in NIFS verify circuit, otherwise error will be catch up in the final CompressSNARK check.

Updated with https://github.com/microsoft/Nova/pull/204/commits/a938f293ddbf8dab782ed012f07cf8616965cd5b
Thanks again for the review :)

src/supernova/circuit.rs
```diff
@@ -0,0 +1,751 @@
...
+    // of its outputs as public output. In the subsequent step, the next augmented
+    // function takes as non-deterministic input a preimage to this hash." pg.16
+
+    // https://eprint.iacr.org/2022/1758.pdf
+
+    // Compute the new hash H(params, i+1, program_counter, z0, z_{i+1}, U_next)
+    let mut ro = G::ROCircuit::new(
+      self.ro_consts.clone(),
+      2 // params_next, i_new
+      + program_counter.as_ref().map(|_| 1).unwrap_or(0) // optional program counter
```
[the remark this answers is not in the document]
hero78119: 👍 updated in https://github.com/microsoft/Nova/pull/204/commits/f76dfc2d6276b50272feaeaae57e0f35dffe5838

src/supernova/circuit.rs
```diff
@@ -0,0 +1,751 @@
...
+    arity: usize,
+    last_augmented_circuit_index: &AllocatedNum<G::Base>,
+    program_counter: Option<AllocatedNum<G::Base>>,
+    num_augmented_circuits: usize,
+  ) -> Result<(AllocatedRelaxedR1CSInstance<G>, AllocatedBit), SynthesisError> {
+    // Check that u.x[0] = Hash(params, i, program_counter, U[], z0, zi)
+    let mut ro = G::ROCircuit::new(
+      self.ro_consts.clone(),
+      2 // params_next, i_new
+      + program_counter.as_ref().map(|_| 1).unwrap_or(0) // optional program counter
```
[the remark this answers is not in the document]
hero78119: 👍 updated in https://github.com/microsoft/Nova/pull/204/commits/f76dfc2d6276b50272feaeaae57e0f35dffe5838

src/supernova/circuit.rs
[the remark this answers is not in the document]
porcuquine: Yes, I think this repeated conditional works. Given the size of the values (15 field elements for the allocated relax instance, I *think*), a multicase will be cheaper at about 6 clauses (i.e. 6 running instances). We don't need to worry about that now, but I think we can create an ergonomic and secure solution for this in general. I'll show my work later, perhaps elsewhere, so we can work through the details. For now, I think doing as you suggest would fairly simply solve the soundness issue.

porcuquine:
Hi all, just a quick note, which I've discussed with @hero78119 and @srinathsetty already. Because this work is going to still require further work on the right interfaces and general polishing, we'd like to move it to our (Lurk Lab's) [Arecibo](https://github.com/lurk-lab/arecibo) Nova fork. That will let us iterate on design, and especially simplify using from [`lurk-rs`](https://github.com/lurk-lab/lurk-rs).

This PR and valuable discussion will remain open so @srinathsetty can review the crypto and circuits, with the goal of merging to Nova under a feature flag. Meanwhile, we will continue active review and development of this work in arecibo. Once better answers for some of the unknowns/questions posed by the current work have been incubated, the plan is to bring those back as PRs to Nova.

srinathsetty:
@hero78119 To reiterate, we intend to merge this PR here in Microsoft/Nova once comments are addressed. I'll do a detailed audit of the circuit code and the entire PR soon. We will merge this code under an optional feature flag "super".

@porcuquine It likely makes sense to wait for this merge to happen before further work elsewhere, to avoid duplication of efforts.

src/supernova/circuit.rs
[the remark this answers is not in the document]
hero78119: Issue follow-up: 

Updated a commit to fix relevant soundness issue https://github.com/microsoft/Nova/pull/204/commits/44758846c44c5d7950fa83554d2a453c8bf2129f

Background: there is existing and related soundness cause by `last_augmented_circuit_index` usage, once malicious prover set it outside range `[0, #num_opcode)`. More explicitly, It will cause no-matching of running instance selection in [the code]( https://github.com/microsoft/Nova/blob/f76dfc2d6276b50272feaeaae57e0f35dffe5838/src/supernova/circuit.rs#L447-L468), which means NO UPDATES of running instance*s* U[..]_i+1.

The solution I adopted is to propagate the processed `last_augmented_circuit_index` through whole circuit. The mindset behind is we already have ad-hoc logic to set  `last_augmented_circuit_index=0` once outside range. Then circuit need to use this new variable to assure `U[last_augmented_circuit_index]_i+1` updated accordingly.

Along with new fix, I also document more on this part https://github.com/microsoft/Nova/blob/44758846c44c5d7950fa83554d2a453c8bf2129f/src/supernova/circuit.rs#L411-L428 cc @srinathsetty if you can give more inputs :)

hero78119:
> Hi all, just a quick note, which I've discussed with @hero78119 and @srinathsetty already. Because this work is going to still require further work on the right interfaces and general polishing, we'd like to move it to our (Lurk Lab's) [Arecibo](https://github.com/lurk-lab/arecibo) Nova fork. That will let us iterate on design, and especially simplify using from [`lurk-rs`](https://github.com/lurk-lab/lurk-rs).
> 
> This PR and valuable discussion will remain open so @srinathsetty can review the crypto and circuits, with the goal of merging to Nova under a feature flag. Meanwhile, we will continue active review and development of this work in arecibo. Once better answers for some of the unknowns/questions posed by the current work have been incubated, the plan is to bring those back as PRs to Nova.

Yes I have already discussed with @porcuquine and @huitseeker,  it makes sense to incubating and polishing more on interface for applications, and (Lurk Lab's) [Arecibo](https://github.com/lurk-lab/arecibo) is nice candidate to explore more on it.
Another application I am exploring is integrate Supernova to [Powdr](https://powdr-labs.github.io/powdr/) and [ongoing PR](https://github.com/powdr-labs/powdr/pull/453) as first folding-based backend. 
With more applications it will help to judging APIs, supported usage patterns etc :)

> @hero78119 To reiterate, we intend to merge this PR here in Microsoft/Nova once comments are addressed. I'll do a detailed audit of the circuit code and the entire PR soon. We will merge this code under an optional feature flag "super".
> 
> @porcuquine It likely makes sense to wait for this merge to happen before further work elsewhere, to avoid duplication of efforts.

Yes it sounds great to me. (+ cc co-author @wyattbenno777 )

hero78119:
> Updated: Optimize supernova RecursiveSNARK prove/verify step, circuit input,... to pass by reference almost everywhere, around 10% improvement based on benchmark.
> 
> Also run [benchmark recursive-snark](https://gist.github.com/hero78119/5e955ddd51135b12a7bf26f54148dd17) for Nova vs Supernova-1-augmented-circuit version In summary, in low constraints case no much different, on high constraints case Supernova is around 5-10% slower. However on largest constraints `1038757` supernova is 10% fast.
> 
> Benchmark is on windows-subsystem-linux so perf will be variant. Will be better to run benchmark on other neutral server to get more accurate result.

Share [benchmark result](https://gist.github.com/hero78119/5e955ddd51135b12a7bf26f54148dd17?permalink_comment_id=4662685#gistcomment-4662685) on latest commit vs old commit

drawing

Summary
- recent few optimisation on Nova: optimise memory/dependency/trait interface ... optimise both Nova/Supernova prover quite a lot, especially on largest constraints `1038757` seems > 4x fast.
- supernova vs nova no significant performance difference after recently few optimisation

srinathsetty:
Based on the offline discussion, @porcuquine and team will submit a new PR that has been improved in their fork of Nova. We will review the PR once it is stable and submitted.

hero78119:
Another review ongoing PR https://github.com/lurk-lab/arecibo/pull/48 which is to unlock the full power of Nova to support read-write lookup in cheap cost, leveraging offline-memory-check tricks mentioned in various paper. It's related to this SuperNova PR where the ROM based on this to achieve constant size of Nova state instead of linear to ROM size

srinathsetty:
We will close this PR given alternate plans to get this back.

PR #205, opened 2023-07-17
test: Tests for polynomial & bignat circuits

huitseeker:
- Used struct `PolynomialMultiplier` with a corresponding test
- Added `proptest` to dependencies for property-based testing
- Created property-based test to verify big number decomposition under various conditions

### DIFF
--- a/src/gadgets/nonnative/bignat.rs
+++ b/src/gadgets/nonnative/bignat.rs
@@ -783,7 +783,9 @@ impl<Scalar: PrimeField> Polynomial<Scalar> {
 #[cfg(test)]
 mod tests {
   use super::*;
-  use bellperson::Circuit;
+  use bellperson::{gadgets::test::TestConstraintSystem, Circuit};
+  use pasta_curves::pallas::Scalar;
+  use proptest::prelude::*;
 
   pub struct PolynomialMultiplier<Scalar: PrimeField> {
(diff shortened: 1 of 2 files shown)


src/gadgets/nonnative/bignat.rs
```diff
@@ -818,4 +820,79 @@ mod tests {
...
+        let circuit = BigNatBitDecomp {
+           inputs: Some(BigNatBitDecompInputs {
+                n: BigInt::from(n),
+            }),
+            params: BigNatBitDecompParams {
+                limb_width: limb_width as usize,
+                n_limbs,
+            },
+        };
+        let mut cs = TestConstraintSystem::<pasta_curves::pallas::Scalar>::new();
```
srinathsetty, line 893: Is it worth making the test generic?

huitseeker: I thought about it, but as:
-  it works on a field, rather than with the group,
- and actually works with relatively small numbers (for performance reasons, note that n is selected as a u16),
I thought it may not be worth it at this stage. 

Let me know if you think otherwise - I intend to go a tad deeper into this section of the code in any case.

srinathsetty: That makes sense, thanks! 

Yes, it will be fantastic to have more eyes on this section!

PR #219, opened 2023-08-11
Simplifications to trait structures

huitseeker:
The individual commit messages are meant to be meaningful. At a high level:

-  starts leveraging `zkcrypto/group` to simplify trait definitions (see #198, but consider this only reuses the blanket impl of a convenience trait referring to `std::ops`), 
- removes unneeded trait bounds (this is backwards-compatible), 
- removes unneeded `Phantom` types,
- updates `halo2curves` with the recently released package,
- refactors `CommitmentKeyExtTrait` to remove unneeded associated type (this is a breaking change, but I've tested [the changes](https://github.com/lurk-lab/lurk-rs/blob/5ddd40f4be4808d7b460688ab90837122cc46eb4/src/proof/nova.rs#L58-L60), which are minimal),
- refactors `EvaluationEngineTrait` to remove unneeded associated type,

### DIFF
--- a/src/provider/bn256_grumpkin.rs
+++ b/src/provider/bn256_grumpkin.rs
@@ -10,13 +10,11 @@ use crate::{
 };
 use digest::{ExtendableOutput, Update};
 use ff::{FromUniformBytes, PrimeField};
+use group::{cofactor::CofactorCurveAffine, Curve, Group as AnotherGroup, GroupEncoding};
 use num_bigint::BigInt;
 use num_traits::Num;
-use pasta_curves::{
-  self,
-  arithmetic::{CurveAffine, CurveExt},
-  group::{cofactor::CofactorCurveAffine, Curve, Group as AnotherGroup, GroupEncoding},
-};
+// Remove this when https://github.com/zcash/pasta_curves/issues/41 resolves
+use pasta_curves::arithmetic::{CurveAffine, CurveExt};
 use rayon::prelude::*;
 use sha3::Shake256;
 use std::io::Read;
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -33,6 +33,7 @@ bitvec = "1.0"
 byteorder = "1.4.3"
 thiserror = "1.0"
 halo2curves = { version = "0.1.0", features = ["derive_serde"] }
+group = "0.13.0"
 
 [target.'cfg(any(target_arch = "x86_64", target_arch = "aarch64"))'.dependencies]
 pasta-msm = { version = "0.1.4" }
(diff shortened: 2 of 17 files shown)


srinathsetty:
Thanks for the PR @huitseeker! As always, this is great!

As discussed, could we please remove dependency on the group crate from the "core" of Nova (e.g., in Nova's traits), at least for the time being?

src/provider/bn256_grumpkin.rs
```diff
@@ -10,13 +10,11 @@ use crate::{
 };
 use digest::{ExtendableOutput, Update};
 use ff::{FromUniformBytes, PrimeField};
+use group::{cofactor::CofactorCurveAffine, Curve, Group as AnotherGroup, GroupEncoding};
 use num_bigint::BigInt;
 use num_traits::Num;
-use pasta_curves::{
```
srinathsetty, line 15: Should we make a similar change in src/provider/pasta.rs?

huitseeker: I do not expect the version of the `::group` traits re-exported in the `::pasta_curves` crate to ever be out of sync w.r.t the traits implemented by the `::pasta_curves` structs. 
The point of the change you’re commenting on is to reflect that the `halo2curves` crate, while using crates from `::group`, is not particularly bound (notably in their versions) to those re-exported by the `pasta_curves` crates, with which it does not have a relationship.

huitseeker:
> As discussed, could we please remove dependency on the group crate from the "core" of Nova (e.g., in Nova's traits), at least for the time being?

For the record of a future reader diving in the Rust aspects of this: there was no dependency on a group trait in any of [ae1440b](https://github.com/microsoft/Nova/pull/219/commits/ae1440ba050d7f0cde83abb22f48fd5530597113): this imported a convenience [blanket trait](https://doc.rust-lang.org/book/ch10-02-traits.html#using-trait-bounds-to-conditionally-implement-methods), for which all the supertraits are defined *outside *of the group crate (in `std::ops`), and which is [not implementable in another way than the provided default](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=4624b2924148a22b32e74e3616829ca6). In other terms, the traits that a user of Nova has to care about are all in `std::ops`, none of them are in `group`.

With that said, it’s a 3-line change, and scanning the imports of the `traits` module and not finding the word `group` in there does get the point across simply.

src/provider/bn256_grumpkin.rs
srinathsetty: Yes, that makes sense, thanks!

PR #220, opened 2023-08-16
Refine comments

Jun-Hee-Lee:
Dropped `let (right, _) = right.split_at(n);` since `MultilinearPolynomial::new ` ensures even length.

### DIFF
--- a/src/spartan/ppsnark.rs
+++ b/src/spartan/ppsnark.rs
@@ -47,8 +47,13 @@ impl<Scalar: PrimeField> IdentityPolynomial<Scalar> {
 
   pub fn evaluate(&self, r: &[Scalar]) -> Scalar {
     assert_eq!(self.ell, r.len());
-    (0..self.ell)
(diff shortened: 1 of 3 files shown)


src/spartan/ppsnark.rs
```diff
@@ -288,7 +293,7 @@ pub trait SumcheckEngine<G: Group> {
   /// the size of the polynomials
   fn size(&self) -> usize;
 
-  /// returns evaluation points at 0, 2, d-1 (where d is the degree of the sum-check polynomial)
+  /// returns evaluation points at 0, 2, 3 (where 3 is the degree of the sum-check polynomial)
```
srinathsetty, line 296: Nit: the degree of the sum-check need not be 3.

Jun-Hee-Lee: Hello. Thank you for the guidance. Then, in my opinion, 0, 2, 3, ..., d seems more appropriate. I have reverted this part. Thank you.

PR #221, opened 2023-08-17
Refactors + testing around Spartan Polynomials

huitseeker:


src/spartan/polys/eq_poly.rs
```diff
@@ -0,0 +1,97 @@
+//! `EqPolynomial`: Represents multilinear extension of equality polynomials, evaluated based on binary input values.
```
srinathsetty, line 1: Perhaps, cleaner to call the file eq.rs?

src/spartan/polys/unipoly.rs
```diff
@@ -0,0 +1,186 @@
+//! Main components:
```
srinathsetty, line 1: univariate.rs?

src/spartan/polys/mlpoly.rs
```diff
@@ -1,80 +1,17 @@
-//! This module provides foundational types and functions for manipulating multilinear polynomials in the context of cryptographic computations.
```
srinathsetty, line 1: multilinear.rs?

src/spartan/polys/mod.rs
```diff
@@ -0,0 +1,4 @@
+//! This module contains the definitions of polynomial types used in the Spartan SNARK.
+pub mod eq;
+pub mod multilinear;
```
srinathsetty, line 3: Do we need to export these outside the crate?

huitseeker: These are currently public:
https://github.com/microsoft/Nova/blob/main/src/lib.rs#L25
https://github.com/microsoft/Nova/blob/main/src/spartan/mod.rs#L10

The answer to your question is no, we don't, except for the IPA, which uses `EqPolynomial`. 
But then the same question could be asked of most types in the `spartan` module, and answered the same, so if we're going to ask ourselves visibility questions, what scope should we consider?

I think that morally speaking, many types in the `EvaluationEngine` suite of traits *should* refer to types components of a multilinear polynomial library, and don't, because the notions of point, polynomial and evaluation are loosely defined in terms of scalars & scalar slices with implicit semantics.

I think it would be a great improvement to define those types cleanly, and then analyze visibility from this basis. The present PR seems a rather fragile place to start by comparison.

PR #222, opened 2023-08-21
Small RO trait simplification 

huitseeker:
- makes RO traits a tad more approachable by removing the `ROConstants` trait, 
- removes `Sized` bounds (on `Group`, which is used as a generic parameters, they are redundant)

This is a backport of https://github.com/lurk-lab/arecibo/pull/7

### DIFF
--- a/src/traits/mod.rs
+++ b/src/traits/mod.rs
@@ -117,14 +117,12 @@ pub trait AbsorbInROTrait<G: Group> {
 }
 
 /// A helper trait that defines the behavior of a hash function that we use as an RO
-pub trait ROTrait<Base, Scalar> {
+pub trait ROTrait<Base: PrimeField, Scalar> {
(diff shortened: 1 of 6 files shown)


src/traits/mod.rs
```diff
@@ -138,8 +139,11 @@ pub trait ROTrait<Base, Scalar> {
 
 /// A helper trait that defines the behavior of a hash function that we use as an RO in the circuit model
 pub trait ROCircuitTrait<Base: PrimeField> {
-  /// A type representing constants/parameters associated with the hash function
-  type Constants: ROConstantsTrait<Base>
+  /// the vanilla alter ego of this trait - this constrains it to use the same constants
+  type VanillaRO<T: PrimeField>: ROTrait<Base, T, Constants = Self::Constants>;
```
srinathsetty, line 143: Minor nit: Would it make sense to call this `NativeRO`?

PR #233, opened 2023-10-14
refactor: remove clippy attributes

bingcicle:
Simple PR getting rid of some clippy attributes.

`clippy::upper_case_acronyms` seems to not trigger for `ShapeCS` and `TestShapeCS`, so this just removes the unnecessary allows for those.

Removing `clippy::needless_range_loop` is a bit more opinionated. Used a `for_each()` to get rid of it but conceptually I can see it being easier for some to read a _for i in X_ notation instead, even though the result is the same. I can undo this change, if the _for i in X_ style is preferable here.

The only clippy attribute that can't be gotten rid of is the `upper_case_acronyms` annotation for `NIFS`.


bingcicle:

srinathsetty:
Looks good! thanks!

PR #234, opened 2023-10-17
refactor: Parametrization of lib.rs Tests for Various Evaluation Engines (Arecibo backport)

huitseeker:
[microsoft/171](https://github.com/microsoft/Nova/pull/171) introduced enhanced genericity in tests to facilitate unit testing across diverse curve cycles.  Since then, the lib.rs tests, being designed to verify the behaviors of SNARKs instantiated for certain sets of parameters, tend to embed the use of a RelaxedR1CSSnark<G, EE>. They do this under the alias (where the test carefully chooses a specific import path for RelaxedR1CSSnark):

```rust
type S<G> = RelaxedR1CSSnark<G, ipa_pc::EvaluationEgine<G>>
```
Effectively, this constrains the PCS utilized in producing a succinct proof of knowledge of a valid IVC proof to the IPA, even if the only thing the test function needed to be specific about is `RelaxedR1CSSnark`.

Hence, these generic tests then had to accommodate constraints facilitating the use of the IPA, specifically:

```
<G::CE as CommitmentEngineTrait<G>>::CommitmentKey: CommitmentKeyExtTrait<G>
```
This is because the IPA operates exclusively with a "splittable" key, a concept delineated by the CommitmentKeyExtTrait.

However, two core adjustments are required:

- For the implementation of different PCSs (than the existing IPA), it's imperative to render `EE<G>: EvaluationEngineTrait<G>` configurable.
- Parametrizing the tests based on the `EvaluationEngineTrait` implementation eliminates boundaries associated with `CommitmentKeyExtTrait`. Such boundaries matching the `EvaluationEngine` and `CommitmentEngine` only materialize when the scheme is instantiated (i.e., during test invocation), a point where they are seamlessly met.

The present PR effects this parametrization, effectively making the lib.rs tests functions variable based on the evaluation engine in use.

Backport of https://github.com/lurk-lab/arecibo/pull/70

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -806,14 +806,14 @@ type CE<G> = <G as Group>::CE;
 #[cfg(test)]
 mod tests {
   use crate::provider::bn256_grumpkin::{bn256, grumpkin};
-  use crate::provider::pedersen::CommitmentKeyExtTrait;
   use crate::provider::secp_secq::{secp256k1, secq256k1};
+  use crate::traits::evaluation::EvaluationEngineTrait;
   use core::fmt::Write;
 
   use super::*;
   type EE<G> = provider::ipa_pc::EvaluationEngine<G>;
-  type S<G1> = spartan::snark::RelaxedR1CSSNARK<G1, EE<G1>>;
-  type SPrime<G1> = spartan::ppsnark::RelaxedR1CSSNARK<G1, EE<G1>>;
+  type S<G, EE> = spartan::snark::RelaxedR1CSSNARK<G, EE>;
+  type SPrime<G, EE> = spartan::ppsnark::RelaxedR1CSSNARK<G, EE>;
 
   use ::bellpepper_core::{num::AllocatedNum, ConstraintSystem, SynthesisError};
   use core::marker::PhantomData;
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -1086,13 +1088,13 @@ mod tests {
...
+  fn test_ivc_nontrivial_with_compression_with<G1, G2, E1, E2>()
   where
     G1: Group<Base = <G2 as Group>::Scalar>,
     G2: Group<Base = <G1 as Group>::Scalar>,
-    // this is due to the reliance on CommitmentKeyExtTrait as a bound in ipa_pc
-    <G1::CE as CommitmentEngineTrait<G1>>::CommitmentKey: CommitmentKeyExtTrait<G1>,
-    <G2::CE as CommitmentEngineTrait<G2>>::CommitmentKey: CommitmentKeyExtTrait<G2>,
+    E1: EvaluationEngineTrait<G1>,
+    E2: EvaluationEngineTrait<G2>,
+    // this is due to the reliance on Abomonation
```
srinathsetty, line 1097: This comment seems to be a leftover. Can we cut this comment about Abomonation?

srinathsetty:
Thanks @huitseeker! I left a tiny comment about the code.

For the record, the following quoted text doesn't quite parse and might be due to a common terminology misconception in this space. Decider is an algorithm that checks if a predicate holds given an instance and an optional witness. Similarly, it is not quite clear what a decider SNARK is.
 
> Effectively, this constrains the PCS utilized by that decider to the IPA

I think what you mean by the above terminology is the notion of producing a succinct proof of knowledge of a valid IVC proof. (This was first described in the Nova paper using Spartan with IPA-PCS and Dory-PCS.) The code refers to this instantiation as `CompressedSNARK`. A decider (introduced in a different paper) does not refer to this.

src/lib.rs
huitseeker: of course!

huitseeker:
> For the record, the following quoted text doesn't quite parse and

Fair enough, I've edited the PR summary.

PR #238, opened 2023-10-27
Digest simplifications

srinathsetty:
PR #229 introduces simplifications to digest computation, but it makes the verifier compute digests on its own, changing the behavior prior to that PR. This PR restores that functionality with an internal type.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "nova-snark"
-version = "0.24.0"
+version = "0.25.0"
 authors = ["Srinath Setty <srinath@microsoft.com>"]
 edition = "2021"
 description = "Recursive zkSNARKs without trusted setup"
--- a/src/r1cs/mod.rs
+++ b/src/r1cs/mod.rs
@@ -5,7 +5,6 @@ mod util;
 
 use crate::{
   constants::{BN_LIMB_WIDTH, BN_N_LIMBS},
-  digest::{DigestComputer, SimpleDigestible},
   errors::NovaError,
   gadgets::{
     nonnative::{bignat::nat_to_limbs, util::f_to_nat},
@@ -18,8 +17,6 @@ use crate::{
 };
 use core::{cmp::max, marker::PhantomData};
 use ff::Field;
-use once_cell::sync::OnceCell;
-
 use rayon::prelude::*;
 use serde::{Deserialize, Serialize};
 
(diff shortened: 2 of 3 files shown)


porcuquine:
I think the digest computations removed in https://github.com/microsoft/Nova/pull/238/commits/9f9d45c4aa8981071268cfba61bd3a3b8fce844d are there so that applications (like Lurk) can generate accurate cache keys for public parameters. In particular, we need to invalidate caches whenever circuits change.

srinathsetty:
> I think the digest computations removed in [9f9d45c](https://github.com/microsoft/Nova/commit/9f9d45c4aa8981071268cfba61bd3a3b8fce844d) are there so that applications (like Lurk) can generate accurate cache keys for public parameters. In particular, we need to invalidate caches whenever circuits change.

R1CSShape is an internal type that isn't exposed, I don't think, to users of Nova. If so, how will Lurk determine when the R1CS digest changes? The public parameters type still has a digest field. Please let me know if I misunderstood the intended usage of digest field on R1CSShape type.

srinathsetty:
> > I think the digest computations removed in [9f9d45c](https://github.com/microsoft/Nova/commit/9f9d45c4aa8981071268cfba61bd3a3b8fce844d) are there so that applications (like Lurk) can generate accurate cache keys for public parameters. In particular, we need to invalidate caches whenever circuits change.
> 
> R1CSShape is an internal type that isn't exposed, I don't think, to users of Nova. If so, how will Lurk determine when the R1CS digest changes? The public parameters type still has a digest field. Please let me know if I misunderstood the intended usage of digest field on R1CSShape type.

Please open an issue and/or a PR (or reply here) if the digest on R1CS is still required.

huitseeker:
## On this PR

> [#229] makes the verifier compute digests on its own

I am not sure that I understand this statement with precision. In #229, as in prior behavior, both `spartan::snark` and `spartan::ppsnark` exit the `setup()` function with a verifier key which digest is computed and populated (that is, the `OnceCell` it contains is set). That is because the digest of the `VerifierKey` needs to be populated in the `Proverkey`. There is no digest re-computation involved past that point: further `.digest()` calls [resolve to a simple getter](https://docs.rs/once_cell/latest/once_cell/sync/struct.OnceCell.html#method.get_or_init).

The difference introduced in #229 is that should the `VerifierKey` in question then be serialized and deserialized, the first access of the `digest()` would need to recompute it. This is so that it's not possible to pass the verifier a struct with a digest that doesn't match the associated contents of the verifier key.

In other terms, in order for this line:
https://github.com/microsoft/Nova/blob/bd6e4e1f6479162e0e6b9c58e535a1197ef93158/src/spartan/ppsnark.rs#L1518-L1519
to involve an actual digest re-computation, the verifier would need to be passed a verifier key that did not go through the `setup` function. 

Are you saying that for the verifier to be passed a `VerifyingKey` which digest works for the purpose of the FS heuristic, but contains a wrong parameter (e.g. a wrong num_cons, an incorrect basis in the evaluation engine's verifier key) is ok? If not, then shouldn't that line involve a digest re-computation anyway?

## On R1CS digests

- the `R1CSShape` type is part of the Nova public API, through e.g. [`RelaxedR1CSSNARKTrait`](https://github.com/microsoft/Nova/blob/71ecb66145b7967a8546011134c56602af60697a/src/traits/snark.rs#L22-L25), 
- by and large, the Nova Public parameters are not universal, due to the embedded `R1CSShape` dependent items it contains. Everything else is: the constants are the same from one proof to another, and the only constrain on `CommitmentKey`s is that they be of sufficient sizes for the proof at hand. Having separate digests for the universal and circuit-specific portions of the logic is useful.

PR #244, opened 2023-10-30
chore: run github actions parallelly

ashWhiteHat:
Hi there

Github actions check not only build but also test, fmt and clippy so I divided into some parts.
It can run CI check parallelly and save time.

**Question**
I am reading about [Revisiting the Nova Proof System on a Cycle of Curves](https://eprint.iacr.org/2023/969.pdf#page=5&zoom=150,150,750).

committed

Is $R1CS^{(2)} := (A_2,B_2,C_2 \in \mathbb F_2^{n_2 \times m_2})$ correct instead of $R1CS^{(2)} := (A_2,B_2,C_2 \in \mathbb F_1^{n_2 \times m_2})$?
Sorry if I am wrong.

Thank you.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -10,27 +10,67 @@ jobs:
   build:
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v2
-    - name: Install
-      run: rustup default stable
-    - name: Install rustfmt Components
-      run: rustup component add rustfmt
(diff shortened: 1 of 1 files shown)


srinathsetty:
Thanks for the CI update, it looks great! 

Regarding your question, yes, that is indeed correct. We will fix the typo.

ashWhiteHat:
Thank you for the review!

PR #249, opened 2023-10-31
Move to actual Group trait

ashWhiteHat:
close https://github.com/microsoft/Nova/issues/198

I moved to Zcash `Group` to reduce duplication of code and reduce the complexity of the lib.

1st: Rename Group -> GroupExt
2nd: Rename AnotherGroup -> Group
3rd: Move to Group trait
4th: Apply changes to benches and examples
5th: Remove GroupExt code duplication

Thank you.

### DIFF
--- a/benches/compressed-snark.rs
+++ b/benches/compressed-snark.rs
@@ -4,11 +4,9 @@ use bellpepper_core::{num::AllocatedNum, ConstraintSystem, SynthesisError};
 use core::marker::PhantomData;
 use criterion::*;
 use ff::PrimeField;
+use group::Group;
 use nova_snark::{
-  traits::{
-    circuit::{StepCircuit, TrivialCircuit},
-    GroupExt,
-  },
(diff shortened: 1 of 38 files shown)


srinathsetty:
Hi @ashWhiteHat, thanks for the PR! We have been thinking about substantially simplifying the traits. In particular, we plan to move to anchor the code around `CommitmentEngine` trait as most of Nova does not quite care about the specific Group operations. It only relies on the ability to commit to vectors and take linear combinations of commitments. We will keep this PR around in case that experiment does not go as planned.

ashWhiteHat:
That's cool.
Thank you for letting me know.
I am going to work on other stuff.

srinathsetty:
Hi @ashWhiteHat, given #263, the Group trait is substantially simplified so we likely won't need this PR. Please let me know if you think this might still be useful. Thanks again for your work!

PR #254, opened 2023-11-07
test: msm test and refactoring

ashWhiteHat:
Hi there

I added msm test.
In field, it has inverse except zero and that error can be handled by `invert()` method Option trait so I integrated it.

Thank you.

### DIFF
--- a/src/provider/mod.rs
+++ b/src/provider/mod.rs
@@ -262,28 +262,41 @@ macro_rules! impl_traits {
 
 #[cfg(test)]
 mod tests {
-  use group::{ff::Field, Group};
-  use rand_core::OsRng;
-
   use super::cpu_best_multiexp;
 
-  type F = pasta_curves::pallas::Scalar;
-  type G = pasta_curves::pallas::Point;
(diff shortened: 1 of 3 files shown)


srinathsetty (approved):

src/provider/mod.rs
```diff
@@ -268,3 +259,31 @@ macro_rules! impl_traits {
...
+  use rand_core::OsRng;
+
+  use super::cpu_best_multiexp;
+
+  type F = pasta_curves::pallas::Scalar;
+  type G = pasta_curves::pallas::Point;
+  type A = pasta_curves::pallas::Affine;
+
+  #[test]
+  fn msm_test() {
```
huitseeker, line 275: It would be useful to make this test generic and operating on all curve pairs supported by the repository (in a way similar to how the other tests in the repo function).

srinathsetty: Good point!

ashWhiteHat: Thank you for the review.
I applied your suggestion.

PR #255, opened 2023-11-08
Reduce duplicate code across different curve cycle providers

ashWhiteHat:
I reduced test and transcript trait code across curve cycle by macros.

**improvement**
curve cycle group methods difference is only vartime_multiscalar_mul.
https://github.com/microsoft/Nova/blob/main/src/provider/mod.rs#L159
If we call msm method through such that `Self::msm()`, we can use same trait between pasta and other cycle pair.

**question**
Is there any reason not to use complete addition for `ecc` gadget?
https://github.com/microsoft/Nova/blob/main/src/gadgets/ecc.rs#L135
We can skip condition branch constraint.

**typo**
`the` duplication
the
https://eprint.iacr.org/2023/1192.pdf#page=4&zoom=100,100,250

Thank you.

### DIFF
--- a/src/provider/mod.rs
+++ b/src/provider/mod.rs
@@ -229,29 +229,14 @@ macro_rules! impl_traits {
     $order_str:literal,
     $base_str:literal
   ) => {
-    impl Engine for $engine {
-      type Base = $name::Base;
-      type Scalar = $name::Scalar;
-      type GE = $name::Point;
-      type RO = PoseidonRO<Self::Base, Self::Scalar>;
-      type ROCircuit = PoseidonROCircuit<Self::Base>;
-      type TE = Keccak256Transcript<Self>;
-      type CE = CommitmentEngine<Self>;
-    }
-
-    impl Group for $name::Point {
-      type Base = $name::Base;
-      type Scalar = $name::Scalar;
-
-      fn group_params() -> (Self::Base, Self::Base, BigInt, BigInt) {
-        let A = $name::Point::a();
-        let B = $name::Point::b();
-        let order = BigInt::from_str_radix($order_str, 16).unwrap();
-        let base = BigInt::from_str_radix($base_str, 16).unwrap();
-
-        (A, B, order, base)
-      }
-    }
+    impl_folding!(
+      $engine,
+      $name,
+      $name_compressed,
+      $name_curve,
+      $order_str,
+      $base_str
+    );
 
     impl DlogGroup for $name::Point {
       type CompressedGroupElement = $name_compressed;
@@ -335,10 +320,11 @@ macro_rules! impl_traits {
       }
     }
 
-    impl PrimeFieldExt for $name::Scalar {
-      fn from_uniform(bytes: &[u8]) -> Self {
-        let bytes_arr: [u8; 64] = bytes.try_into().unwrap();
-        $name::Scalar::from_uniform_bytes(&bytes_arr)
+    impl CompressedGroup for $name_compressed {
+      type GroupElement = $name::Point;
+
+      fn decompress(&self) -> Option<$name::Point> {
+        Some($name_curve::from_bytes(&self).unwrap())
       }
     }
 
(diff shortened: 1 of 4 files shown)


src/traits/mod.rs
```diff
@@ -45,7 +46,13 @@ pub trait Group:
   type CompressedGroupElement: CompressedGroup<GroupElement = Self>;
 
   /// A type representing preprocessed group element
-  type PreprocessedGroupElement: Clone + Debug + Send + Sync + Serialize + for<'de> Deserialize<'de>;
+  type PreprocessedGroupElement: CurveAffine
```
huitseeker, line 49: This is probably worth linking to #198

srinathsetty: Hi, @ashWhiteHat could we avoid adding `CurveAffine` requirement to this type here? This introduces a dependency on group crate from any curve library that can be used to implement a provider. What is the gain from adding this requirement?

ashWhiteHat: Thank you for the question.
That's necessary for test abstraction of [test_from_label](https://github.com/microsoft/Nova/blob/main/src/provider/bn256_grumpkin.rs#L92).

It checks return value of [from_label](https://github.com/microsoft/Nova/blob/main/src/provider/bn256_grumpkin.rs#L97) and [from_label_serial](https://github.com/microsoft/Nova/blob/main/src/provider/bn256_grumpkin.rs#L98) on [assert_eq](https://github.com/microsoft/Nova/blob/main/src/provider/bn256_grumpkin.rs#L101).

Thus, return value of **from_label** and **from_label_serial** should be the same type.
In **from_label_serial**, we use [`hash_to_curve`](https://github.com/microsoft/Nova/blob/main/src/provider/bn256_grumpkin.rs#L85) and [`to_affine`](https://github.com/microsoft/Nova/blob/main/src/provider/bn256_grumpkin.rs#L86).
**hash_to_curve** is method of CurveExt of [pasta](https://github.com/zcash/pasta_curves/blob/main/src/arithmetic/curves.rs#L67) and **to_affine** converts the type of CurveAffine.

To match the type, it was necessary.

I may be able to do it with macros and avoid CurveAffine requirement.

ashWhiteHat: Yes, I generalized the test without changing trait.
I would appreciate it if you could check.

src/provider/mod.rs
```diff
@@ -347,12 +333,48 @@ macro_rules! impl_traits {
...
       }
     }
+  };
+}
 
-    impl CompressedGroup for $name_compressed {
-      type GroupElement = $name::Point;
+/// Nova folding circuit engine and curve group ops
+#[macro_export]
+macro_rules! impl_folding {
```
srinathsetty, line 341: Is there a reason to call this `impl_folding`? It seems more appropriate to call it `impl_engine`?

ashWhiteHat: That makes more sense to me.
I renamed accordingly.

srinathsetty:
> close #202
> 
> I reduced test and transcript trait code across curve cycle by macros.

Does it actually close the issue? There is still impl_traits that is duplicated between mod.rs and pasta.rs, right?

srinathsetty:
> question
> Is there any reason not to use complete addition for ecc gadget?
> https://github.com/microsoft/Nova/blob/main/src/gadgets/ecc.rs#L135
> We can skip condition branch constraint.

I don't fully understand this question. Can you please elaborate?

ashWhiteHat:
> Does it actually close the issue? There is still impl_traits that is duplicated between mod.rs and pasta.rs, right?

Exactly.
I unlinked the issue.
> I don't fully understand this question. Can you please elaborate?

Sorry for confusing.
This seems my mistake.
I thought that the curve complete addition law can save the number of constraints because it can skip condition branch (identity and doubling case).
but it can save the number of constraints because it can perform affine coordinate arithmetic by using the feature that the inversion cost is as cheap as multiplication as following article says.
https://bitzecbzc.github.io/technology/jubjub/index.html#introducing-jubjub

It seems we can't save the number of constraints, because the Weierstrass affine coordinate still needs condition branch.

PR #263, opened 2023-11-16
Refactor traits that allows implementing different engines for the same curve cycle

srinathsetty:
Anchors code around `Engine` trait rather than `Group` trait so one can build different implementations of the `Engine` trait for the same curve cycle.

Building on #259, allows customizing `Group` trait to the particulars of a commitment/evaluation argument.


src/lib.rs
```diff
@@ -121,63 +121,63 @@ where
...
   /// # use nova_snark::provider::ipa_pc::EvaluationEngine;
-  /// # use nova_snark::traits::{circuit::TrivialCircuit, Group, snark::RelaxedR1CSSNARKTrait};
+  /// # use nova_snark::traits::{circuit::TrivialCircuit, Engine, snark::RelaxedR1CSSNARKTrait};
   /// use nova_snark::PublicParams;
   ///
-  /// type G1 = pallas::Point;
-  /// type G2 = vesta::Point;
-  /// type EE<G> = EvaluationEngine<G>;
-  /// type SPrime<G> = RelaxedR1CSSNARK<G, EE<G>>;
+  /// type E1 = pallas::Point;
```
srinathsetty, line 127: This document comment needs to be fixed.

PR #264, opened 2023-11-17
Small improvement to the minroot example

field-worker:
About 10% improvement for the non-release mode. My guess is that examples are typically no run with the `release` flag.


srinathsetty (approved):
Thanks!

PR #281, opened 2023-12-10
test: multilinear polynomial evaluation endianness (Arecibo backport)

huitseeker:
This back-ports the following PR from Arecibo : https://github.com/lurk-lab/arecibo/pull/128

The endianness of evaluation-domain points in Nova is sometimes surprising (there are [examples in other repos](https://github.com/privacy-scaling-explorations/Nova/issues/20)), so we found useful to document it in lavish detail with this test.


src/spartan/polys/multilinear.rs
```diff
@@ -287,4 +289,119 @@ mod tests {
...
+        .collect();
+      // this shows the order in which coordinates are evaluated
+      let rev_pt: Vec<_> = pt.iter().cloned().rev().collect();
+      assert_eq!(poly.evaluate(&pt), partial_eval(&poly, &rev_pt).Z[0])
+    }
+  }
+
+  #[test]
+  fn test_partial_and_evaluate() {
+    partial_and_evaluate_with::<Fp>();
```
srinathsetty, line 403: For uniformity, could we use pasta_curves::Fp here?

src/spartan/polys/multilinear.rs
```diff
@@ -287,4 +289,119 @@ mod tests {
...
+    mut rng: &mut R,
+  ) -> MultilinearPolynomial<Scalar> {
+    MultilinearPolynomial::new(
+      std::iter::from_fn(|| Some(Scalar::random(&mut rng)))
+        .take(1 << num_vars)
+        .collect(),
+    )
+  }
+
+  /// This evaluates a multilinear polynomial at a partial point in the evaluation domain,
```
srinathsetty, line 305: Can you elaborate a bit more on the purpose of this test? What exactly are we testing here? I'm also not sure what it means to evaluate a multilinear polynomial at a "partial" point? I don't think that notion is well defined or standard.

huitseeker: I've simplified the test, re-formulated to take advantage of `bind_poly_var_top`.
I believe this not doesn't introduce any notion that wasn't already there, please take a look.

srinathsetty: Thanks!

src/spartan/polys/multilinear.rs
```diff
@@ -283,8 +284,65 @@ mod tests {
...
+    assert!(poly.Z.len() >= 1 << values.len());
+
+    let mut tmp = poly.clone();
+    for v in values.iter() {
+      tmp.bind_poly_var_top(v);
+    }
+    tmp
+  }
+
+  fn partial_and_evaluate_with<F: PrimeField>() {
```
srinathsetty, line 326: nit: `partial` --> `bind` for improved consistency?

huitseeker: Renamed to the two functions which output is compared in the test.

srinathsetty: thanks!

PR #286, opened 2024-01-01
Improve error handling

jbearer:
* When a function already returns a `Result`, propagate errors instead of panicking with `expect`
* For `NovaError::SynthesisError`, retain information about the original bellpepper error. Since `NovaError` implements `Clone` but bellpepper's `SynthesisError` does not, we keep the error information as a `String`.

This commit only fixes low-hanging fruit in lib.rs, for functions that already return a `Result` and can easily propagate errors just by replacing `expect(...)` with `?`. There are still many `unwrap()` calls in functions returning `Result` in other modules, particularly gadgets. But I don't understand the code well in those parts, and I suspect some of those `unwrap()`s actually can't fail based on invariants of the code, so it makes perfect sense to leave them as is.

### DIFF
--- a/src/errors.rs
+++ b/src/errors.rs
@@ -58,12 +58,23 @@ pub enum NovaError {
   #[error("IncorrectWitness")]
   IncorrectWitness,
   /// return when error during synthesis
-  #[error("SynthesisError")]
-  SynthesisError,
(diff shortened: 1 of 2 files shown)


jbearer:

Issue #284, opened 2023-12-20
Remove redundant absorbing to the transcript


storojs72:
One of `CompressedSNARK`'s verification steps (affects only [ppspartan.rs](https://github.com/microsoft/Nova/blob/main/src/spartan/ppsnark.rs)) - checking the satisfiability of the folded instances ([link](https://github.com/microsoft/Nova/blob/main/src/lib.rs#L833)) - has [following](https://github.com/microsoft/Nova/blob/main/src/spartan/ppsnark.rs#L1367) redundant absorption of `eval_vec` to the transcript, since it is already happened [here](https://github.com/microsoft/Nova/blob/main/src/spartan/ppsnark.rs#L1346). Absorbing the same value twice adds no new entropy to the transcript, so second one can be freely removed.

I'm happy to provide a hotfix for this in case if core contributors confirm the issue

srinathsetty:
Hi @storojs72, thanks the report! Yes, they are indeed absorbed twice (this is not a correctness issue, but a cost that does not need to be paid). Yes, please submit a fix.

PR #287, opened 2024-01-03
chore: Remove redundant absorption to the transcript

storojs72:
Fixes https://github.com/microsoft/Nova/issues/284

### DIFF
--- a/src/spartan/ppsnark.rs
+++ b/src/spartan/ppsnark.rs
@@ -1050,20 +1050,16 @@ impl<E: Engine, EE: EvaluationEngineTrait<E>> RelaxedR1CSSNARKTrait<E> for Relax
     let (comm_L_row, comm_L_col) =
       rayon::join(|| E::CE::commit(ck, &L_row), || E::CE::commit(ck, &L_col));
 
-    // absorb the claimed evaluations into the transcript
-    transcript.absorb(
-      b"e",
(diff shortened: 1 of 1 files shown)


storojs72:
@microsoft-github-policy-service agree company="LurkLab"

srinathsetty:
Hi! Thanks for the PR! I'm a little confused about the diff. The simplest fix for the issue is to just remove the following two lines. Isn't that cleaner than what is proposed here? I'm not entirely sure why there should be any other edits (e.g., regarding the absorption of comm_L_row and comm_L_col).

Prover: https://github.com/microsoft/Nova/blob/fe5b93234ec9b972dde2d0406fbe2d1e9984acfc/src/spartan/ppsnark.rs#L1066
Verifier: https://github.com/microsoft/Nova/blob/fe5b93234ec9b972dde2d0406fbe2d1e9984acfc/src/spartan/ppsnark.rs#L1367

storojs72:
Right. But, let's consider verification flow. Why do we use [this](https://github.com/microsoft/Nova/blob/main/src/spartan/ppsnark.rs#L1348) manually crafted slice, if [later](https://github.com/microsoft/Nova/blob/main/src/spartan/ppsnark.rs#L1361) we create `eval_vec` with exactly same elements? So I thought that moving absorptions without violation of their order (`&eval_vec.as_slice()` first, and `&vec![comm_L_row, comm_L_col].as_slice` second) after introducing `eval_vec` would be more readable.

Issue #56, opened 2022-05-13
Optimization: The running instance does not need to absorbed.


srinathsetty:
Given that the public IO of the incoming R1CS instance contains the hash of the running instance in addition to (params, U, i, z0, zi), we can skip absorbing the running instance.

varunthakore:
Hi, can I work on this issue ?

My understanding is that since u.x[0] = Hash(params, U, i, z0, zi) and we are absorbing `u` at line 241, we can avoid absorbing `U` at line 240. 
https://github.com/microsoft/Nova/blob/48065b0c708fc836b6f2db316b0170b3a4e05292/src/gadgets/r1cs.rs#L240-L241

Similarly in NIFS::prove and NIFS::verify, we can avoid absorbing the RelaxedR1CSInstance `U1`.
https://github.com/microsoft/Nova/blob/48065b0c708fc836b6f2db316b0170b3a4e05292/src/nifs.rs#L49
https://github.com/microsoft/Nova/blob/48065b0c708fc836b6f2db316b0170b3a4e05292/src/nifs.rs#L95

We will have to update the const `NUM_FE_FOR_RO`  and expected values in tests accordingly.

Let me know if this is fine. I can create a PR.

srinathsetty:
@varunthakore Yes, that sounds great!

PR #291, opened 2024-01-09
Remove absorbing of running instance

varunthakore:
This PR is corresponding to issue #56. It removes absorbing of running instance and updates the value of `NUM_FE_FOR_RO` and expected values in tests.


varunthakore:

src/gadgets/r1cs.rs
```diff
@@ -237,7 +237,6 @@ impl<E: Engine> AllocatedRelaxedR1CSInstance<E> {
     // Compute r:
     let mut ro = E::ROCircuit::new(ro_consts, NUM_FE_FOR_RO);
     ro.absorb(params);
-    self.absorb_in_ro(cs.namespace(|| "absorb running instance"), &mut ro)?;
```
srinathsetty, line 240: Could we add a note here that we are not absorbing the running instance because "fill in"?

PR #283, opened 2023-12-18
feat: implement SuperNova

huitseeker:
This implements [Supernova](https://eprint.iacr.org/2022/1758), allowing:
- a 'pay-as-you-go' cost structure for folding operations, through the SuperNova folding scheme,
- a final SNARK that efficiently compresses an instance of this folded proof, through batching techniques.

References:
- the [blog post](https://blog.lurk-lang.org/posts/arecibo-supernova/#technical-release-note-supernova-protocol-integration-into-nova) goes into our construction, and links where relevant to two more specialized notes on [the `CompressedSNARK` for Supernova](https://hackmd.io/@adr1anh/BJw1g0aBT) along with our variant of the [public input padding issue](https://hackmd.io/@adr1anh/Sy08YaVBa).
- the Readme at `src/supernova/Readme.md`

This backports the following Arecibo PRs:
- https://github.com/lurk-lab/arecibo/pull/2
- https://github.com/lurk-lab/arecibo/pull/3
- https://github.com/lurk-lab/arecibo/pull/10
- https://github.com/lurk-lab/arecibo/pull/16
- https://github.com/lurk-lab/arecibo/pull/23
- https://github.com/lurk-lab/arecibo/pull/30
- https://github.com/lurk-lab/arecibo/pull/28
- https://github.com/lurk-lab/arecibo/pull/41
- https://github.com/lurk-lab/arecibo/pull/45
- https://github.com/lurk-lab/arecibo/pull/50
- https://github.com/lurk-lab/arecibo/pull/56
- https://github.com/lurk-lab/arecibo/pull/51
- https://github.com/lurk-lab/arecibo/pull/72
- https://github.com/lurk-lab/arecibo/pull/92
- https://github.com/lurk-lab/arecibo/pull/95
- https://github.com/lurk-lab/arecibo/pull/97
- https://github.com/lurk-lab/arecibo/pull/101
- https://github.com/lurk-lab/arecibo/pull/110
- https://github.com/lurk-lab/arecibo/pull/106
- https://github.com/lurk-lab/arecibo/pull/112
- https://github.com/lurk-lab/arecibo/pull/114
- https://github.com/lurk-lab/arecibo/pull/119
- https://github.com/lurk-lab/arecibo/pull/120
- https://github.com/lurk-lab/arecibo/pull/127
- https://github.com/lurk-lab/arecibo/pull/123
- https://github.com/lurk-lab/arecibo/pull/131
- https://github.com/lurk-lab/arecibo/pull/174
[...]

PR #293, opened 2024-01-17
refactor: Improve Spartan SNARK polynomial computations and evaluations

huitseeker:
## Summary 

This gathers the changes to the pre-processing SNARK excerpted from the Supernova implementation in PR #283. The main changes extracted here are the introduction of the masked eq polynomial (and its use fixing [this issue](https://hackmd.io/@adr1anh/Sy08YaVBa)), additional sum-check tooling, removal of two calls to evaluation argument.

Main reference Arecibo PRs:
- https://github.com/lurk-lab/arecibo/pull/106
- https://github.com/lurk-lab/arecibo/pull/131
- https://github.com/lurk-lab/arecibo/pull/174
- https://github.com/lurk-lab/arecibo/pull/182

<details>
<summary> More details (click to expand) </summary>

- Enhancement of polynomial related code in Spartan protocol including new polynomial types, modified existing types, better evaluation methods, and improved polynomial operations.
- Introduction of `squares` function and change in the generation of `t_pow` in Spartan.
- Addition of a new polynomial type through `MaskedEqPolynomial` with methods for its creation and evaluation.
- Enhancements in `UniPoly` struct by addition of `PartialEq`, and `Eq` traits.
- Improvements in `snark.rs` for proving and verifying batch evaluation claims, leveraging `gamma` and `rho` for random linear combinations, and optimizing various variable computations.
- Updates in `multilinear.rs` with refactoring, optimization, error handling, and supplementing unit tests.
- Refactor in `spartan/mod.rs` with import updates, function overhauls, struct visibility changes, and asynchronous operations for efficient calculations.
- Additions and amendments in `sumcheck.rs` for batch verification, handling of vectors of claims, handling of cubic bounds with additive terms, visibility adjustments, and typo fixes.
- Modifications in `eq.rs` including a debug derive for `EqPolynomial`, enhanced visibility of `r` vector, provision of `evals_from_points` for enhanced evaluation, and addition of `FromIterator` implementation.
</details>

### DIFF
--- a/src/spartan/mod.rs
+++ b/src/spartan/mod.rs
@@ -5,24 +5,29 @@
 //! We also provide direct.rs that allows proving a step circuit directly with either of the two SNARKs.
 //!
 //! In polynomial.rs we also provide foundational types and functions for manipulating multilinear polynomials.
+pub mod direct;
 #[macro_use]
 mod macros;
-pub mod direct;
 pub(crate) mod math;
 pub mod polys;
 pub mod ppsnark;
 pub mod snark;
 mod sumcheck;
 
-use crate::{traits::Engine, Commitment};
+use crate::{
+  r1cs::{R1CSShape, SparseMatrix},
+  traits::Engine,
+  Commitment,
+};
 use ff::Field;
 use itertools::Itertools as _;
 use polys::multilinear::SparsePolynomial;
 use rayon::{iter::IntoParallelRefIterator, prelude::*};
 
+// Creates a vector of the first `n` powers of `s`.
 fn powers<E: Engine>(s: &E::Scalar, n: usize) -> Vec<E::Scalar> {
   assert!(n >= 1);
-  let mut powers = Vec::new();
+  let mut powers = Vec::with_capacity(n);
   powers.push(E::Scalar::ONE);
   for i in 1..n {
     powers.push(powers[i - 1] * s);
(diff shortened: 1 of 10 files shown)


src/spartan/mod.rs
```diff
@@ -84,25 +114,54 @@ impl<E: Engine> PolyEvalWitness<E> {
...
+    x: Vec<E::Scalar>,
+    s: E::Scalar,
+  ) -> PolyEvalInstance<E> {
+    let num_instances = num_vars.len();
+    assert_eq!(c_vec.len(), num_instances);
+    assert_eq!(e_vec.len(), num_instances);
+
+    let num_vars_max = x.len();
+    let powers: Vec<E::Scalar> = powers::<E>(&s, num_instances);
+    // Rescale evaluations by the first Lagrange polynomial,
```
srinathsetty, line 137: I'm a bit confused about this rescaling. This wasn't necessary in the prior version (https://github.com/microsoft/Nova/blob/8244d00a7b3bf5fc44e8cf76c3a0d67cfdb17394/src/spartan/mod.rs#L109) didn't need this rescaling. Can you elaborate what changed?

adr1anh: During batched Sumcheck over `n` variables, we a interpret polynomial `P` with `m<n` as having `n-m` variables inserted "in front" (the original `m` variables are `X_{n-m}, ..., X_{n-1}`). This means that Sumcheck will return the evaluation `P(r_{n-m}, ..., r_{n-1})`. However, we are using a PCS that supports polynomials in `n` variables, and we committed to `P` by doing an MSM with the first `2^m` entries of the commitment key. This means we really committed to the polynomial `P'(X_0, ..., X_{n-1}) = L_0(X_0, ..., X_{n-m-1}) * P(X_{n-m}, ..., X_{n-1})`, where only the first `2^m` entries are non-zero. 

The degree bounding checks for `P` are all implicitly performed by the Sumcheck instances, since we use evaluation of `eq` with the appropriate number of variables, all computed by the verifier. The scaling that happens in `batch` is just a way of allowing for the two interpretations `P` and `P'`.

srinathsetty: Hi Adrian! Thanks for the response! I understood the scaling part (in a way, this is similar to the scaling of IO-related polynomials when combining W and IO to account for padding, right?).

What is still confusing me is that we didn't need to do such scaling in the prior version (linked above). Can you elaborate why we need scaling of evaluations here but we didn't need it before?

adr1anh: That's right, the idea is essentially the same as the `W` and `IO` vectors. 

In the prior version, all vectors were padded with zeros to the same `N = 2^n`. A polynomial `P` that only had `m < n` variables would be interpreted from the start as `L_0(X_0, ..., X_{n-m-1})P(X_{n-m}, ..., X_{n-1})`, including during Sumcheck. Therefore, this scaling is applied by Sumcheck during the `bind` phase. In the first `n-m` rounds, the second half of the polynomial will be 0, so the effect is just a multiplication of the vector by `(1-r_i)`. 

The batched Sumcheck handles the unpadded `P` and only evaluates it in `r_{n-m}, ..., r_{n-1}`, avoiding the multiplication by `(1-r_i)` in the first `n-m` rounds. 

Essentially, padding with zeros will force Sumcheck to do the scaling. Since we avoid the padding, we need to manually add the scaling by the Lagrange polynomials in the batch open phase.

PR #295, opened 2024-01-17
move fold step from CompressedSNARK to RecursiveSNARK

srinathsetty:
Nova paper's description performs a fold step prior to applying a SNARK. Revisiting Nova paper observes that this fold step does not alter the incrementality property. 

This PR moves the fold step from CompressedSNARK to RecrursiveSNARK, so the RecursiveSNARK needs to track a single witness rather than two witnesses for the secondary circuit.


srinathsetty:
Closing this for now.

PR #306, opened 2024-02-05
feat: make r1cs compoents public

RyanKung:
For now, mod r1cs and it's components are private, this PR aims to make them public, thus can be used outside nova repo.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -41,7 +41,7 @@ use errors::NovaError;
 use ff::Field;
 use gadgets::utils::scalar_as_base;
 use nifs::NIFS;
-use r1cs::{
+pub use r1cs::{
   CommitmentKeyHint, R1CSInstance, R1CSShape, R1CSWitness, RelaxedR1CSInstance, RelaxedR1CSWitness,
 };
 use serde::{Deserialize, Serialize};


RyanKung:
@microsoft-github-policy-service agree [company="{Rings Network}"]

RyanKung:
@microsoft-github-policy-service agree company="rings network"

srinathsetty:
Hi @RyanKung Could you clarify the need to use R1CS-related types through Nova's APIs?

A key principle in Nova is to minimize public APIs so users of Nova work with a simple API. In that vein, the R1CS module is expected to be internal-only types. So, can you please clarify your use case?

RyanKung:
> Hi @RyanKung Could you clarify the need to use R1CS-related types through Nova's APIs?
> 
> A key principle in Nova is to minimize public APIs so users of Nova work with a simple API. In that vein, the R1CS module is expected to be internal-only types. So, can you please clarify your use case?

I hope to implement my own fold algorithm externally, so I hope Nova's private modules can have appropriate permissions. If this does not align with Nova's design purpose, it's OK to close this PR.

srinathsetty:
> > Hi @RyanKung Could you clarify the need to use R1CS-related types through Nova's APIs?
> > A key principle in Nova is to minimize public APIs so users of Nova work with a simple API. In that vein, the R1CS module is expected to be internal-only types. So, can you please clarify your use case?
> 
> I hope to implement my own fold algorithm externally, so I hope Nova's private modules can have appropriate permissions. If this does not align with Nova's design purpose, it's OK to close this PR.

Even if we make r1cs module public, it is not clear it will help you build your own folding scheme. The repo is not parameterized by the folding scheme yet.

PR #307, opened 2024-02-05
support serialize and deserialize for engines

RyanKung:


RyanKung:
@microsoft-github-policy-service agree [company="{Rings Network}"]

RyanKung:
@microsoft-github-policy-service agree company="rings network"

srinathsetty:
Hi @RyanKung Could you please clarify the need for serde on Engine related types?

RyanKung:
> Hi @RyanKung Could you please clarify the need for serde on Engine related types?

My use case is like

```rust
use crate::provider::VestaEngine;
use crate::provider::PallasEngine;

#[derive(Serialize, Deserialize, Debug, Clone)]
pub struct Wrap<E1, E2>
where
      E1: Engine<Base = <E2 as Engine>::Scalar>,
      E2: Engine<Base = <E1 as Engine>::Scalar>,
{
    inner: (E1, E2),
}

#[derive(Serialize, Deserialize, Clone)]
enum Example
{
    VestaPallas(Wrap<VestaEngine, PallasEngine>),
    PallasVasta(Wrap<PallasEngine, VestaEngine>),
}
```

This will cause a compile error:

```shell
error[E0277]: the trait bound `VestaEngine: Serialize` is not satisfied
   --> src/lib.rs:855:10
    |
855 | #[derive(Serialize, Deserialize, Clone)]
    |          ^^^^^^^^^ the trait `Serialize` is not implemented for `VestaEngine`
...
858 |     VestaPallas(Wrap<VestaEngine, PallasEngine>),
    |                 ------------------------------- required by a bound introduced by this call
    |
    = help: the following other types implement trait `Serialize`:
              bool
              char
              isize
              i8
              i16
              i32
              i64
              i128
            and 238 others
```

src/provider/mod.rs
```diff
@@ -26,13 +26,15 @@ use crate::{
   traits::Engine,
 };
 use pasta_curves::{pallas, vesta};
+use serde::Deserialize;
```
srinathsetty, line 29: Nit: use serde::{Deserialize, Serialize};

src/lib.rs
```diff
@@ -53,7 +53,7 @@ use traits::{
 };
 
 /// A type that holds public parameters of Nova
-#[derive(Serialize, Deserialize)]
+#[derive(Serialize, Deserialize, Clone)]
```
srinathsetty, line 56: Why do we need Clone on public parameters? This type is always passed by reference, so clone seems unnecessary.

src/provider/mod.rs
RyanKung: Thanks, just updated.

PR #315, opened 2024-03-07
feat: Refactor R1CS shape to split the commitment key generation

huitseeker:
This is a fragment of #283 

## TL;DR: 

Splits off one major source of diff lines from PR #283. Inherently, there are many R1CS shapes to consider when tailoring public parameter creation to non-uniform step-circuits. However, the commitment key generation should only be done once for all circuits (once a suitable size has been determined by looking at all R1CS shapes). This splits the relevant Nova functions into `r1cs_shape_and_key`, `r1cs_shape` and `commitment_key` to enable the flexibility deamnded by the above model.

## In detail:
- Renamed the `r1cs_shape` method across various files to `r1cs_shape_and_key`, indicating its functionality is to return both `R1CSShape` and `CommitmentKey`.
- Altered function calls from `r1cs_shape` to `r1cs_shape_and_key` in files such as `direct.rs`, `nifs.rs`, `lib.rs` and `circuit.rs`,
- Split the creation of `R1CSShape` and `CommitmentKey` into separate functions in the `NovaShape` object in `r1cs.rs`
- Removed the `R1CS` struct in `mod.rs` as it only contained a phantom data, with related operations performed elsewhere.
- Implemented changes to enhance code readability, including the addition of a new `commitment_key_size` function, and overall code reformatting for clarity.

### DIFF
--- a/src/bellpepper/mod.rs
+++ b/src/bellpepper/mod.rs
@@ -45,7 +45,7 @@ mod tests {
     // First create the shape
     let mut cs: ShapeCS<E> = ShapeCS::new();
     synthesize_alloc_bit(&mut cs);
-    let (shape, ck) = cs.r1cs_shape(&*default_ck_hint());
+    let (shape, ck) = cs.r1cs_shape_and_key(&*default_ck_hint());
 
     // Now get the assignment
     let mut cs = SatisfyingAssignment::<E>::new();
--- a/src/bellpepper/mod.rs
+++ b/src/bellpepper/mod.rs
@@ -45,7 +45,7 @@ mod tests {
     // First create the shape
     let mut cs: ShapeCS<E> = ShapeCS::new();
     synthesize_alloc_bit(&mut cs);
-    let (shape, ck) = cs.r1cs_shape_and_key(&*default_ck_hint());
+    let (shape, ck) = cs.r1cs_shape_and_ck(&*default_ck_hint());
 
     // Now get the assignment
     let mut cs = SatisfyingAssignment::<E>::new();
--- a/src/r1cs/mod.rs
+++ b/src/r1cs/mod.rs
@@ -481,7 +481,7 @@ impl<E: Engine> RelaxedR1CSInstance<E> {
     let mut r_instance = RelaxedR1CSInstance::default(ck, S);
     r_instance.comm_W = instance.comm_W;
     r_instance.u = E::Scalar::ONE;
-    r_instance.X = instance.X.clone();
+    r_instance.X.clone_from(&instance.X);
     r_instance
   }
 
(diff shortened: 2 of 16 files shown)


src/bellpepper/mod.rs
```diff
@@ -45,7 +45,7 @@ mod tests {
     // First create the shape
     let mut cs: ShapeCS<E> = ShapeCS::new();
     synthesize_alloc_bit(&mut cs);
-    let (shape, ck) = cs.r1cs_shape(&*default_ck_hint());
+    let (shape, ck) = cs.r1cs_shape_and_key(&*default_ck_hint());
```
srinathsetty, line 48: Nit: rename `r1cs_shape_and_key` --> `r1cs_shape_and_ck` for consistency?

src/r1cs/mod.rs
```diff
@@ -77,24 +70,31 @@ pub struct RelaxedR1CSInstance<E: Engine> {
...
+pub fn commitment_key<E: Engine>(
+  S: &R1CSShape<E>,
+  ck_floor: &CommitmentKeyHint<E>,
+) -> CommitmentKey<E> {
+  let size = commitment_key_size(S, ck_floor);
+  E::CE::setup(b"ck", size)
+}
+
+/// Computes the number of generators required for the commitment key corresponding to shape `S`.
+fn commitment_key_size<E: Engine>(S: &R1CSShape<E>, ck_floor: &CommitmentKeyHint<E>) -> usize {
```
srinathsetty, line 93: Does it make sense to pass a vector/slice of R1CSShape objects to `commitment_key` method and have it pick the right size? This will avoid having two methods and logic above this layer?

huitseeker: You can check the use of `commitment_key_size` in #283 in the `src/supernova/mod.rs`, in `compute_primary_ck`. This is left for a further PR so as to make those further changes in supernova-specific files, according to the general thrust of your [comments](https://github.com/microsoft/Nova/pull/283#issuecomment-1875777020) in #283.

srinathsetty: Thanks for the clarification! I think we could still have `commitment_key` to be a method inside R1CSShape rather than having two separate methods. For now, it takes (S: &R1CSShape<E>, ck_floor: &CommitmentKeyHint<E>) as arguments and returns a ck. In the future, it can take a collection of R1CSShape and hints and returns a ck. 

In other words, I don't see what is gained from having two methods, one called `commitment_key` and another called `commitment_key_size`. Why not inline the contents of `commitment_key_size` inside `commitment_key` method and then move `commitment_key` to be a method under `R1CSShape` type rather than a stand-alone method.

srinathsetty:
Since this PR is stale, closing it for now. Please reopen if you wish to contribute.

Issue #174, opened 2023-05-30
Zero-knowledge implementation gap


huitseeker:
As outlined in the [Lurk audit report](https://blog.lurk-lang.org/posts/alpha-audit/inference-security-assessment-3-2023.pdf), the variants of the [IPA](https://github.com/microsoft/Nova/commit/cddd707fada6b2778f7d6d0667b8c47e3eb43d56) and [Spartan](https://github.com/microsoft/Nova/commit/54f758eef36a42bec189c9f27f153fd7545f6ac3) in the repository do not provide zero-knowledge (unlike the sum-check in the original [Spartan repository](https://github.com/microsoft/Spartan/commit/e0f964c3f93dc95134ba36940b22e0ad6b09fe98)).

In particular, the IPA needs to be updated to use blinding commitments and the sum-check inside Spartan needs to be made hiding. 

At a more abstract level, [Hyrax](https://eprint.iacr.org/2017/1132) offers a blueprint on how to make the sum-check zero-knowledge.

huitseeker:
https://github.com/microsoft/Spartan2/pull/1 Looks super nice!

ekrembal:
Hello,

I'm wondering about the current status of this issue. Has there been any progress on this issue, and is there a timeline for addressing it?

Thank you for the update.

srinathsetty:
This one is addressed in #330

PR #330, opened 2024-10-02
Zero Knowledge Fold for Nova IVC from HyperNova

jkwoods:
An implementation of the zero-knowledge fold from Appendix D.4 of the HyperNova paper.

Changes "from the bottom up":

`src/r1cs/mod.rs`:
- new `fold_relaxed()` methods for `RelaxedR1CSInstance` and `RelaxedR1CSWitness` (Previously, we were only able to `fold()` RelaxedR1CSInstances/Witnesses with regular non-relaxed R1CSInstance/Witnesses. Since the random layer is a Relaxed instance, we need to be able to fold two RelaxedR1CSInstances/Witnesses together.)
- new `commit_T_relaxed()` method. Again, the same as `commit_T()`, but for two RelaxedR1CSInstances/Witnesses. Takes error terms into account.
- new `sample_random_instance_witness()` method to sample a random RelaxedR1CSInstance/Witness pair. Uses `OsRng` for now. Based on Construction 5 from Appendix D in the HyperNova paper.
- new test for `sample_random_instance_witness()`

`src/nifs.rs`:
- new `prove_relaxed()` and `verify_relaxed()` that are similar to their `prove()` and `verify()` counterparts, except for two RelaxedR1CSInstances/Witnesses. They used the `fold_relaxed()` and `commit_T_relaxed()` methods from `src/r1cs/mod.rs`. They use a new constant, `NUM_FE_FOR_RO_RELAXED`, created in `src/constants.rs`, since we have to absorb more into the random oracle.
- new methods in the test module for testing relaxed proving and verifying and folding in a random instance.

`src/lib.rs`:
- new testing functions for the random IVC layer

For `RecursiveSNARK`:
- new `randomizing_fold()` method. This is like `prove_step()` except instead of proving a new F iteration, it folds in a random layer. It returns a new `RandomLayerRecursive` struct with information about the random fold, rather than modifying the `RecursiveSNARK` struct. `randomizing_fold()` is public right now (for completeness), but shouldn’t be used if the user wants to compress their snark, as calling `CompressedSNARK`’s `prove_randomizing()` will call `randomizing_fold()` itself. If there is a better way to make this tension less confusing, let me know.
- new `verify_randomizing()` method. This method exists for completeness, but probably wouldn't ever be used, as the randomizing layer is the last IVC fold, and will therefore be verified by the `CompressedSNARK` verifier.

For `CompressedSNARK`:
- the `CompressedSNARK` struct as a new optional `RandomLayerCompressed` field. This holds information about the last random fold.
- new `prove_randomizing()` method. the old `prove()` method is renamed to be `prove_regular()` and we make a new `prove()` function that automatically selects the correct prove (regular or randomizing), based on a boolean switch. This boolean switch is the only change that would need to be made to existing code that calls the Nova library. However, if such a change would cause issues, I could restructure this code to be randomizing or not based on a rust feature flag (or something else).
- new `verify_randomizing()` method. Similar to proving, the old `verify()` is renamed to `verify_regular()` and a new `verify()` function automatically selects the correct one, based on whether the `CompressedSNARK` struct has the `RandomLayerCompressed` field.

`benches/compressed-snark.rs`:
- new benchmarks for a compressed snark with a randomizing layer


jkwoods:
@microsoft-github-policy-service agree company="University of Pennsylvania"

srinathsetty:
Hi @jkwoods this is great! thanks!

I'll take a deeper look at the PR. There's however something important missing: the commitments need to be hiding (e.g., they must include a blind at the time of commitment). Otherwise, we won't get the desired hiding property from this transformation.

jkwoods:
@srinathsetty Oh, of course, my bad. I'll get right on that. I assume the blinds also need to be optional/"toggleable".

srinathsetty:
Sounds good! I think it will likely be easier to make blindfolding the default.


On Oct 3, 2024, at 4:17 PM, Jess Woods ***@***.***> wrote:

﻿

@srinathsetty<https://github.com/srinathsetty> Oh, of course, my bad. I'll get right on that. I assume the blinds also need to be optional/"toggleable".

jkwoods:
@srinathsetty It seems straightforward to me to add blinds to the Pedersen commitments, and keep track of how they are folded in the NIFS. But I am a little lost on what happens once the folding is done and we want to compress the IVC proof with a non zero knowledge SNARK. I assume we don't want use have blinded commitments in the spartan SNARK (or the IPA). So is there a point (after the folding in of the random/ZK layer in the IVC?) where we "de-blind" the commitments the Relaxed R1CS instance (by revealing $h^r$?)

jkwoods:
^ Above code adds the blinding into the (Pedersen) commitments. `(Relaxed)R1CSWitness`es save their blinds and fold them in accordance with HyperNova instructions. `CompressedSNARK` proving/verifying will now "derandomize" the `RelaxedR1CSInstance/Witness` pairs after the randomized fold, before the snark. Since `(Relaxed)R1CSInstance/Witness`es are blinded by default now, regular proving/verifying (without the random fold), and the `DirectSNARK` proving/verifying also have to derandomize their  `RelaxedR1csInstance/Witness` pairs.

On the commitment level, even though Pedersen `CommitmentKeys` are now always set up with the $h$ term, the user has the option to `commit()` to something or `commit_with_blinding()`. The r1cs level uses `commit_with_blinding()`, while the rest of the (non-blinded) code uses regular `commit()`. Note for HyperKZG, `commit_with_blinding()` is the same as `commit()`, and the new `derandomize()` method doesn't do anything. (Changing the `CommitmentTrait` required addressing HyperKZG.)

Note the Pedersen `CommitmentKeys` $h$ term is stored in an Option, so that `reinterpret_commitments_as_ck()` can be handled without strange assumptions. Let me know if you had a more elegant implementation in mind for any of this.

Two things still to deal with:
1. extra hashing - I haven't done this yet
2. `test_pp_digest` now fails. I suspect this is because I added the $h$ field to the Pedersen `CommitmentKeys`. Is there a "safe" way to recalculate those expected digest values, or is it as straightforward as just running the new `PublicParams` (with new `CommitmentKeys`) through the `DigestComputer` and getting the new result?

jkwoods:
^ In the above code, randomness is added to the hashing (in circuit and in the clear for the verifier). It is not perfectly clear to me why we add this randomness (so it may need to be calculated in a different way if I have misunderstood); @srinathsetty could you remind me why this is important? Also, I believe all this is finished (except for the `test_pp_digest` mentioned above) - please let me know if anything needs to be changed for either correctness or clarity/organization.

I believe this PR, when it is done, fixes #174.

srinathsetty:
> ^ In the above code, randomness is added to the hashing (in circuit and in the clear for the verifier). It is not perfectly clear to me why we add this randomness (so it may need to be calculated in a different way if I have misunderstood); @srinathsetty could you remind me why this is important?

The random values are added inside a hash to make them simulataable. We can see this invoked in appendix D here: https://eprint.iacr.org/2021/370.pdf

> 
> Also, I believe all this is finished now (except for the `test_pp_digest` mentioned above) - please let me know if anything needs to be changed for either correctness or clarity/organization.

When you run the test, it will tell you the expected value, we can just copy the value test failure prints. This should make the test pass.

> 
> I believe this PR, when it is done, fixes #174.

src/nifs.rs
```diff
@@ -67,7 +70,48 @@ impl<E: Engine> NIFS<E> {
...
+    U2: &RelaxedR1CSInstance<E>,
+    W2: &RelaxedR1CSWitness<E>,
+  ) -> Result<(NIFS<E>, (RelaxedR1CSInstance<E>, RelaxedR1CSWitness<E>)), NovaError> {
+    // initialize a new RO
+    let mut ro = E::RO::new(ro_consts.clone(), NUM_FE_FOR_RO_RELAXED);
+
+    // append the digest of pp to the transcript
+    ro.absorb(scalar_as_base::<E>(*pp_digest));
+
+    // append U2 to transcript, U1 does not need to absorbed since U2.X[0] = Hash(params, U1, i, z0, zi)
```
srinathsetty, line 96: Is this true here? I think `prove_relaxed` gets called with a running instance and a random instance. Since the random instance is sampled uniformly at random, I don't think this comment is applicable. This would mean that we should absorb both instances into the transcript. Isn't it?

src/lib.rs
```diff
@@ -467,10 +509,88 @@ where
...
+    self.ri_primary = r_next_primary;
+    self.ri_secondary = r_next_secondary;
+
     Ok(())
   }
 
-  /// Verify the correctness of the `RecursiveSNARK`
+  /// add a randomizing fold to a `RecursiveSNARK`
+  /// if you plan on using a randomized CompressedSNARK, you don't need to call this
+  pub fn randomizing_fold(
```
srinathsetty, line 520: Does this need to be public?

src/lib.rs
```diff
@@ -712,40 +1006,125 @@ where
...
+      primary_blind_r_E: recursive_snark.r_W_primary.r_E,
+      secondary_blind_r_W: f_W_secondary.r_W,
+      secondary_blind_r_E: f_W_secondary.r_E,
+      random_layer: None,
+      _p: Default::default(),
+    })
+  }
+
+  /// Create a new `CompressedSNARK` (with randomizing layer)
+  fn prove_randomizing(
```
srinathsetty, line 1062: I think given that adding zk is a negligible/tiny cost to the prover and the verifier and proof sizes, I think we should make it the default behavior. This will completely avoid a bunch of code. That is, we could just "upgrade" the existing CompressedSNARK type to provide the zero-knowledge property rather than introducing new methods that one will have to use.

srinathsetty: What do you think @jkwoods?

src/traits/commitment.rs
```diff
@@ -51,8 +51,22 @@ pub trait CommitmentEngineTrait<E: Engine>: Clone + Send + Sync {
...
 
   /// Samples a new commitment key of a specified size
-  fn setup(label: &'static [u8], n: usize) -> Self::CommitmentKey;
+  fn setup(label: &'static [u8], blinding_label: &'static [u8], n: usize) -> Self::CommitmentKey;
 
   /// Commits to the provided vector using the provided generators
   fn commit(ck: &Self::CommitmentKey, v: &[E::Scalar]) -> Self::Commitment;
+
+  /// Commits to the provided vector using the provided generators and random blind
+  fn commit_with_blinding(
```
srinathsetty, line 60: If we make "blinding" the default, we can just update the commit method to accept a blind as an additional input, which will avoid this variant of commit method.

srinathsetty:
Thanks @jkwoods for the PR! I left some comments.

src/lib.rs
jkwoods: This seems good to me! I'll edit.

src/lib.rs
jkwoods: I think probably not. There's no use case where a user might want _just_ the randomized (uncompressed) IVC, right?

src/traits/commitment.rs
jkwoods: That seems good. `src/spartan/snark` and `src/spartan/ppsnark` both need to call an unblinded commit. I can just have then call the new commit method, with 0 as the blind, if that sounds alright?

src/nifs.rs
jkwoods: yep, you're right. I'll edit.

src/traits/commitment.rs
srinathsetty: Yes, exactly, this sounds perfect!

src/lib.rs
srinathsetty: There may be some use-case in the future where one generates a randomized IVC proof on the client-side and sends it to a server for compression. Perhaps, for this PR, we can ignore this use-case to keep the main interface exposed to the outside simple and small.

benches/compressed-snark.rs
```diff
@@ -43,7 +43,7 @@ cfg_if::cfg_if! {
     criterion_group! {
       name = compressed_snark;
       config = Criterion::default().warm_up_time(Duration::from_millis(3000));
-      targets = bench_compressed_snark, bench_compressed_snark_with_computational_commitments
+    targets = bench_compressed_snark, bench_compressed_snark_with_computational_commitments,
```
srinathsetty, line 46: Nit: this space seems to be removed, right? I'm surprised `cargo fmt` doesn't seem to catch this.

src/circuit.rs
```diff
@@ -53,6 +53,8 @@ pub struct NovaAugmentedCircuitInputs<E: Engine> {
   z0: Vec<E::Base>,
   zi: Option<Vec<E::Base>>,
   U: Option<RelaxedR1CSInstance<E>>,
+  ri: Option<E::Base>,
+  r_next: E::Base,
```
srinathsetty, line 57: Does this need to be Option?

src/lib.rs
```diff
@@ -256,6 +259,28 @@ where
   _p: PhantomData<(C1, C2)>,
 }
 
+/// Final randomized fold of `RecursiveSNARK`
+#[derive(Clone, Debug, Serialize, Deserialize)]
+#[serde(bound = "")]
+pub struct RandomLayerRecursive<E1, E2>
```
srinathsetty, line 265: Nit: should we call this `RecursiveSNARKRandomized`?

src/lib.rs
```diff
@@ -256,6 +259,28 @@ where
...
+{
+  last_i: usize,
+  l_ur_primary: RelaxedR1CSInstance<E1>,
+  l_ur_secondary: RelaxedR1CSInstance<E2>,
+  nifs_Uf_secondary: NIFS<E2>,
+  nifs_Un_primary: NIFS<E1>,
+  nifs_Un_secondary: NIFS<E2>,
+  r_Wn_primary: RelaxedR1CSWitness<E1>,
+  r_Wn_secondary: RelaxedR1CSWitness<E2>,
+  // needed for CompressedSNARK proving,
```
srinathsetty, line 278: I'm confused a bit by this comment. These instances are required to verify the correctness of this type as well---regardless of compression right?

src/lib.rs
```diff
@@ -467,10 +509,87 @@ where
...
 
+    self.ri_primary = r_next_primary;
+    self.ri_secondary = r_next_secondary;
+
     Ok(())
   }
 
-  /// Verify the correctness of the `RecursiveSNARK`
+  /// add a randomizing fold to a `RecursiveSNARK`
+  fn randomizing_fold(
```
srinathsetty, line 519: I wonder if this method could be inside `impl RecursiveSNARKRandomized` (if we apply the renaming per https://github.com/microsoft/Nova/pull/330/files#r1806864934)

src/lib.rs
```diff
@@ -813,26 +967,54 @@ where
...
+    // fold secondary U/W with secondary u/w to get Uf/Wf
+    let r_Uf_secondary = self.nifs_Uf_secondary.verify(
       &vk.ro_consts_secondary,
       &scalar_as_base::<E1>(vk.pp_digest),
       &self.r_U_secondary,
       &self.l_u_secondary,
     )?;
 
+    // fold Uf/Wf with random inst/wit to get U1/W1
+    let r_Un_secondary = self.nifs_Un_secondary.verify_relaxed(
```
srinathsetty, line 979: There seems to be a bit of inconsistency here. The code that corresponds to this in the prover is factored out into a separate method, but here, it is inlined right?

src/nifs.rs
```diff
@@ -67,7 +70,52 @@ impl<E: Engine> NIFS<E> {
...
     // fold the witness using `r` and `T`
-    let W = W1.fold(W2, &T, &r)?;
+    let W = W1.fold(W2, &T, &r_T, &r)?;
+
+    // return the folded instance and witness
+    Ok((Self { comm_T }, (U, W)))
+  }
+
+  /// Same as `prove`, but takes two Relaxed R1CS Instance/Witness pairs
+  pub fn prove_relaxed(
```
srinathsetty, line 80: Nit: It might be easier to introduce `NIFSRelaxed` type and have `prove` and `verify` methods on that type. It will eliminate someone from calling `verify_relaxed` when the prover has called `prove` and vice versa. What do you think?

src/provider/hyperkzg.rs
```diff
@@ -185,7 +185,7 @@ where
   type Commitment = Commitment<E>;
   type CommitmentKey = CommitmentKey<E>;
 
-  fn setup(_label: &'static [u8], n: usize) -> Self::CommitmentKey {
+  fn setup(_label: &'static [u8], _blinding_label: &'static [u8], n: usize) -> Self::CommitmentKey {
```
srinathsetty, line 188: Nit: Do we need a separate label for blind generation?

src/r1cs/mod.rs
```diff
@@ -510,6 +659,70 @@ impl<E: Engine> RelaxedR1CSInstance<E> {
...
+
+    RelaxedR1CSInstance {
+      comm_W,
+      comm_E,
+      X,
+      u,
+    }
+  }
+
+  pub fn derandomize_commits(
```
srinathsetty, line 692: Nit: I think we should have a method called `derandomize` in `RelaxedR1CSWitness` and `RelaxedR1CSInstance` rather than have then separately, isn't it?

The prover can call derandomize on witness types first to get back blinds. Then it can pass those blinds to derandomize method on the instances to get back derandomized instances.

This will make the code far cleaner and move these methods to the right types. Isn't it?

benches/compressed-snark.rs
jkwoods: oh, weird. will fix.

src/lib.rs
jkwoods: right now, this is _only_ the random information. The CompressedSNARK prover needs access to the original RecursiveSNARK _and_ this random layer. I could make a larger `RecursiveSNARKRandomized` structure that includes both maybe? (Does that make sense?)

src/lib.rs
jkwoods: Yes, you're right, inconsistent - an artifact of having separate functionality for RecursiveSNARK to prove and verify it's own random layer (before we removed that). I should either separate the first part of verification into a separate function, or combine the proving.

The downside of separating out verification is that it is not a "full verification" (i.e. we will not call `is_sat` at the end, as we would have if we were just verifying IVC proof). (Proving is "full".) I guess this doesn't really matter, but is a different sort of inconsistency.

But separating out functionality is nice and modular ... what do you prefer?

src/nifs.rs
jkwoods: Sounds like a good idea! will do.

src/r1cs/mod.rs
jkwoods: This sounds more elegant, will do.

src/lib.rs
jkwoods: yeah - maybe this method is what makes a `RecursiveSNARKRandomized` object?

src/lib.rs
jkwoods: maybe if both functions are in `RecursiveSNARKRandomized`, which is private to the user (only called inside of CompressedSNARK methods), it doesn't matter if it's not full verification.

src/lib.rs
jkwoods: If you were just verifying a randomized RecursiveSNARK, you would "refold" to get these instances. see [here](https://github.com/microsoft/Nova/blob/2a5cee6f222c651511faa4ddbc4746de57c4e0d2/src/lib.rs#L979) and [here](https://github.com/microsoft/Nova/blob/2a5cee6f222c651511faa4ddbc4746de57c4e0d2/src/lib.rs#L987)

src/circuit.rs
jkwoods: In the base case, we will not have a first r_0, right?

src/provider/hyperkzg.rs
jkwoods: No, we don't. I was trying to create the h generator with a separate call to `from_label` - no reason to do that. I'll change it.

src/lib.rs
srinathsetty: > Yes, you're right, inconsistent - an artifact of having separate functionality for RecursiveSNARK to prove and verify it's own random layer (before we removed that). I should either separate the first part of verification into a separate function, or combine the proving.
> 
> The downside of separating out verification is that it is not a "full verification" (i.e. we will not call `is_sat` at the end, as we would have if we were just verifying IVC proof). (Proving is "full".) I guess this doesn't really matter, but is a different sort of inconsistency.
> 
> But separating out functionality is nice and modular ... what do you prefer?

Since we are not anyway allowing RecursiveSNARK objects that are randomized to be externally used. Perhaps, for this PR, we could just inline the randomization process inside CompressedSNARK's prove method (analogous to how the verify method verifies the randomization). This would also avoid maintaining an additional type right?

src/nifs.rs
srinathsetty: I think it still makes sense to derandomize outside NIFSRelaxed, to avoid confusion.

(Also, absorbing both U1 and U2 in the RO is a standard version, not specialized to the case of a randomized instance. Isn't it?)

src/lib.rs
jkwoods: yes, that sounds good.

src/nifs.rs
jkwoods: will do.

I guess I just meant it doesn't incorporate the previous optimization that ignores the U1 absorption, since U2.X[0] = Hash(params, U1, i, z0, zi).

src/lib.rs
```diff
@@ -623,6 +648,8 @@ where
   pp_digest: E1::Scalar,
   vk_primary: S1::VerifierKey,
   vk_secondary: S2::VerifierKey,
+  vk_primary_commitment_key: CommitmentKey<E1>,
```
srinathsetty, line 651: I missed this earlier. This will make the verifier key non-succinct. Here's a way to fix this.

We should add a type called `DerandKey` (analogous to `CommitmentKey`) in `CommitmentEngine`:

`  /// Holds the type of the derandomization key
  type DerandKey: Clone + Debug + Send + Sync + Serialize + for<'de> Deserialize<'de>;`

We can then add a method like `derand_key` to `CommitmentEngine` and this can be implemented as follows:

`fn derand_key(ck: &Self::CommitmentKey) -> Self::DerandKey {
    DerandKey { h: ck.h.clone() }
  }
`

Then, the verifier only needs to hold a DerandKey and the `derandomize` method only takes this DerandKey, which is constant-sized, making it succinct.

Does this make sense?

src/lib.rs
```diff
@@ -813,26 +920,54 @@ where
...
+      &vk.ro_consts_primary,
+      &vk.pp_digest,
+      &self.r_U_primary,
+      &self.l_ur_primary,
+    )?;
+
+    // derandomize/unblind commitments
+    let derandom_r_Un_primary = r_Un_primary.derandomize(
+      &vk.vk_primary_commitment_key,
+      &self.primary_blind_r_W,
```
srinathsetty, line 950: Nit: `blind_r_W_primary` rather than `primary_blind_r_W` to be consistent with the rest of the names?

Similar comment for other instances.

src/provider/hyperkzg.rs
```diff
@@ -208,12 +208,21 @@ where
...
-  fn commit(ck: &Self::CommitmentKey, v: &[E::Scalar]) -> Self::Commitment {
+  fn commit(ck: &Self::CommitmentKey, v: &[E::Scalar], _r: &E::Scalar) -> Self::Commitment {
     assert!(ck.ck.len() >= v.len());
+
     Commitment {
       comm: E::GE::vartime_multiscalar_mul(v, &ck.ck[..v.len()]),
     }
   }
+
+  fn derandomize(
```
srinathsetty, line 219: Why is this derandomize not using `_r`?

src/provider/hyperkzg.rs
```diff
@@ -208,12 +208,21 @@ where
     Self::CommitmentKey { ck, tau_H }
   }
 
-  fn commit(ck: &Self::CommitmentKey, v: &[E::Scalar]) -> Self::Commitment {
+  fn commit(ck: &Self::CommitmentKey, v: &[E::Scalar], _r: &E::Scalar) -> Self::Commitment {
```
srinathsetty, line 211: Why is this commitment not using `_r`?

src/provider/pedersen.rs
```diff
@@ -167,15 +168,51 @@ where
...
+    } else {
+      assert_eq!(*r, E::Scalar::ZERO);
+
+      Commitment {
+        comm: E::GE::vartime_multiscalar_mul(v, &ck.ck[..v.len()]),
+      }
+    }
+  }
+
+  fn derandomize(
```
srinathsetty, line 202: I'm not sure this method needs to be this complicated and involve MSMs. Why not do something like:

`fn derandomize(dk: &Self::DerandKey, C: &Self::Commitment, blind: E::Scalar) -> Self::Commitment {
    Commitment {
      comm: C.comm - <E::GE as DlogGroup>::group(&dk.h) * blind,
    }
  }`

src/lib.rs
jkwoods: This does make sense! Thank you for all of your help and patience with this PR.

src/lib.rs
```diff
@@ -623,6 +648,8 @@ where
   pp_digest: E1::Scalar,
   vk_primary: S1::VerifierKey,
   vk_secondary: S2::VerifierKey,
+  vk_primary_derand_key: DerandKey<E1>,
```
srinathsetty, line 651: Nit: rename to `dk_primary` and `dk_secondary`

src/lib.rs
```diff
@@ -688,20 +728,24 @@ where
...
+      vk_primary_derand_key: E1::CE::derand_key(&pp.ck_primary),
+      vk_secondary_derand_key: E2::CE::derand_key(&pp.ck_secondary),
       _p: Default::default(),
     };
 
     Ok((pk, vk))
   }
 
-  /// Create a new `CompressedSNARK`
+  /// Create a new `CompressedSNARK` (with randomizing layer)
```
srinathsetty, line 739: Nit: May be just say `(provides zero-knowledge)`

src/nifs.rs
```diff
@@ -213,43 +311,265 @@ mod tests {
     W2: &R1CSWitness<E>,
   ) {
     // produce a default running instance
-    let mut r_W = RelaxedR1CSWitness::default(shape);
-    let mut r_U = RelaxedR1CSInstance::default(ck, shape);
+    let mut running_W = RelaxedR1CSWitness::default(shape);
```
srinathsetty, line 314: Nit: Any reason to rename these?

src/provider/hyperkzg.rs
```diff
@@ -51,6 +52,16 @@ where
   }
 }
 
+/// A type that holds blinding generator
+#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
+pub struct DerandKey<E>
+where
+  E: Engine,
```
srinathsetty, line 59: Nit: E: Engine can be inlined on line 57 as: `pub struct DerandKey <E: Engine>`

srinathsetty:
Hi @jkwoods thanks for the new updates! It looks great! 

I think we are almost near the finish line. I only had some minor comments, which I noted.

src/nifs.rs
jkwoods: `r_W` is used other places to denote the blind for the witness vector. It was just a clarity thing. If you think its better the other way, I can change it back.

srinathsetty: Good point! Let's keep the change.

srinathsetty (approved):

PR #344, opened 2024-12-19
Experimental support for no_std

srinathsetty:


JayWhite2357:
You'll want to check against a true `no_std` target.
Something like this: `cargo check --target thumbv7em-none-eabi --no-default-features`.
Unfortunately, `#![no_std]` is not enough.

srinathsetty:
Close as it is being worked in PR #376

PR #346, opened 2024-12-21
R1CS Folding with ZeroFold

Forpee:
Hi @srinathsetty,

This PR implements the folding scheme for R1CS from the NeutronNova paper.
The full protocol has been implemented.

~The improvements to be addressed are:~
- [x] ~Batch the two Sumfold instances into one, as specified in the paper.~
- [x] ~In the NIFS protocol replace the `Keccak` transcript with `PoseidonRO`, enabling the implementation of a verifier circuit. (I used `keccak` for the initial review becuase the main annoyance I encountered is that the `RO` only allows one squeeze per` RO` instantiation. I would appreciate your input on whether we can modify this behavior.)~
- [x] ~Refactor the `Sumfold` prover implementation, but I'll wait to batch the Sumfold instances first. A `SumFoldEngine` trait may be useful here.~

__

The test is in this file `src/neutron_nova/tests`

```bash
cargo test --lib -- neutron_nova::tests::test_tiny_r1cs_bellpepper
```

Thanks!


srinathsetty:
> In the NIFS protocol replace the Keccak transcript with PoseidonRO, enabling the implementation of a verifier circuit. (I used keccak for the initial review becuase the main annoyance I encountered is that the RO only allows one squeeze per RO instantiation. I would appreciate your input on whether we can modify this behavior.)

Yes, the current RO interfaces squeezing only one challenge. We could potentially expand it to support multiple challenges, but this might make things more complex than necessary. An alternative would be to to create a new transcript, put the last challenge from the previous round as the first entry in the new transcript. This will allow multiple rounds and challenges.

Forpee:
> Yes, the current RO interfaces squeezing only one challenge. We could potentially expand it to support multiple challenges, but this might make things more complex than necessary. An alternative would be to to create a new transcript, put the last challenge from the previous round as the first entry in the new transcript. This will allow multiple rounds and challenges.

@srinathsetty All done! Thank you so much for your expertise!

PR #352, opened 2025-01-10
feat: allow for more general transcripts

JayWhite2357:
# Rationale

HyperKZG requires using `TranscriptReprTrait` for serialization before putting a value on the transcript. This serialization is dependent on the scalar, rather than the transcript.

This become challenging when trying to duplicate the transcript in the EVM, which is big-endian. Allowing for a custom serialization allows for more flexible usage upstream.


JayWhite2357:
Closing in favor of improvements to #353.

PR #358, opened 2025-01-22
refactor: simplify hyperkzg verify

JayWhite2357:
# Rationale

One of the first steps that we did when writing a hyperkzg EVM verifier was simplifying the existing Rust verifier to make a port more direct. The code has been written. It doesn't impact anything other than readability. (With the possible exception of the final commit.)

# Changes

See the individual commits. This PR may be hard to review as a whole, but the individual commits are intended to be very simple.
* The first two commits involve inlining the `kzg_verify_batch` closure, which involves a large `git diff` but very little actual change.
* The remaining commits do various thing. In particular, the main focus is to eliminate "helper" vectors, superfluous variables, mutability, and copy/clone. These all simplify code in general, but in particular make an EVM port much easier.
* The last commit, in particular, is a potential security fix. I believe that the protocol is sound as is, but it is non-trivial to prove this.

### DIFF
--- a/src/provider/hyperkzg.rs
+++ b/src/provider/hyperkzg.rs
@@ -14,14 +14,13 @@ use crate::{
     evaluation::EvaluationEngineTrait,
     AbsorbInROTrait, Engine, ROTrait, TranscriptEngineTrait, TranscriptReprTrait,
   },
-  zip_with,
 };
 use core::{
+  iter,
   marker::PhantomData,
   ops::{Add, Mul, MulAssign},
 };
 use ff::Field;
-use itertools::Itertools;
 use rand_core::OsRng;
 use rayon::prelude::*;
 use serde::{Deserialize, Serialize};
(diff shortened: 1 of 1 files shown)


src/provider/hyperkzg.rs
```diff
@@ -751,7 +736,7 @@ mod tests {
       .with_fixint_encoding()
       .serialize(&proof)
       .unwrap();
-    assert_eq!(proof_bytes.len(), 368);
+    assert_eq!(proof_bytes.len(), 352);
```
srinathsetty, line 739: Quick q: how did the proof length change?

JayWhite2357: Because a `Vec`s length is unknown, `bincode` adds the length of the `Vec` as a `u64` before the actual content of the data. Since this PR turned two `Vec` into fixed-sized arrays, that's two lengths that no longer need to be serialized. Removing `2\times8=16` bytes.

srinathsetty (approved):

PR #366, opened 2025-02-28
feat: add batch commit to trait

jacobtrombetta:


srinathsetty (approved):

src/provider/hyperkzg.rs
```diff
@@ -278,6 +278,24 @@ where
     }
   }
 
+  fn batch_commit(
+    ck: &Self::CommitmentKey,
+    v: &[Vec<<E as Engine>::Scalar>],
+    r: &<E as Engine>::Scalar,
```
srinathsetty, line 284: Apologies for not flagging this earlier. We are commiting to multiple vectors, but all of them are using the same blinding_factor. Why is this okay?

jacobtrombetta: The motivation for a single blinding factor simply came from the current implementations of a batch of commitments in the HyperKZG module - [here](https://github.com/microsoft/Nova/blob/b2f2bd6027856cc5bdcd27d434f0cab7a4eed7f5/src/provider/hyperkzg.rs#L553-L556). Specifically from this code section:
```
let com: Vec<G1Affine<E>> = (1..polys.len())
   .into_par_iter()
   .map(|i| E::CE::commit(ck, &polys[i], &E::Scalar::ZERO).comm.affine())
   .collect();
```
This was likely too much of a simplification on my part. I can update the code to accept multiple blinding_factors.

jacobtrombetta: Thank you for catching that and the explanation. I've updated the `batch_commit` function to require a slice of `r` values to be used as blinding factors.

PR #374, opened 2025-03-11
Add GPU acceleration to bn256_grumpkin trait's vartime and batch vartime multiscalar multiplication

jacobtrombetta:
# Rationale for this change
The [Blitzar project](https://crates.io/crates/blitzar) provides a powerful framework for constructing GPU-accelerated zero-knowledge proofs. A key advantage of Blitzar is its seamless support for various Nvidia GPU models, as well as single and multi-GPU configurations, all without requiring user configuration. The vartime multiscalar multiplication and batched vartime multiscalar multiplication performance defined inside the `bn256_grumpkin` trait can be improved by incorporating the Blitzar crate. 

This PR is related to issue #371.

# What changes are included in this PR? 
- The `blitzar` crate is added as an optional dependency to the project
- A feature flag, `blitzar`, is added
- A `blitzar` module is added to handle `vartime_multiscalar_mul` and `batch_vartime_multiscalar_mul` using the Blitzar crate in this commit: https://github.com/microsoft/Nova/commit/1d7d0f355f85e7cb55ef9eb17e04c26fe0d68b8c
- Blitzar's GPU acceleration is added under the `blitzar` feature flag to the `bn256_grumpkin` trait in this commit: https://github.com/microsoft/Nova/commit/08243cbac533089a283b55c46d05e9a0c7f6407a
- Tests are added to the `blitzar` module


src/provider/bn256_grumpkin.rs
```diff
@@ -30,13 +30,126 @@ pub mod grumpkin {
...
+    )
+    .unwrap();
+    (A, B, order, base)
+  }
+}
+impl DlogGroup for bn256::Point {
+  type AffineGroupElement = bn256::Affine;
+  #[cfg(not(feature = "blitzar"))]
+  fn vartime_multiscalar_mul(scalars: &[Self::Scalar], bases: &[Self::AffineGroupElement]) -> Self {
+    msm_best(scalars, bases)
```
srinathsetty, line 56: There was a recent change to avoid calling into halo2curves' msm library (please see the main branch's impl_traits macro). Could we make this change here?

jacobtrombetta: Thank you both for the review and help. The branch has been rebased and `msm_generic` has been added.

src/provider/bn256_grumpkin.rs
```diff
@@ -33,13 +33,123 @@ pub mod grumpkin {
...
+        let mut shake = Shake256::default();
+        shake.update(label);
+        let mut reader = shake.finalize_xof();
+        let mut uniform_bytes_vec = Vec::new();
+        for _ in 0..n {
+            let mut uniform_bytes = [0u8;
+            32];
+            reader.read_exact(&mut uniform_bytes).unwrap();
+            uniform_bytes_vec.push(uniform_bytes);
+        }let gens_proj:Vec<Bn256Point>  = (0..n).into_par_iter().map(|i|{
```
srinathsetty, line 84: Is there some issue with `cargo fmt` here? It looks like `let gens_proj` is inlined with `}`

srinathsetty:
> > > In Nova, we would run MSMs with scalars represented with integer types (like u8, u16, u32, etc.). Please see the commit bench for details of this. Does blitzar have APIs for this?
> > 
> > 
> > Yep. We didn't integrate it into this PR since it wasn't obvious how to do this outside of the benches, and doing it in just the benches seems less-than-useful.
> 
> There's now a `commit_small` method inside the `CommitmentEngineTrait` (and the corresponding lower-level method in DlogGroupTrait. This should make it possible to integrate, right?
> 
> > > Also, a bunch of this code is unrelated to calling into blitzar or Nova-local MSM code on CPUs. Is there a way to factor out that code so we don't have duplicate copies of this code in BN254-specific implementation?
> > 
> > 
> > It doesn't seem like there is an easy way without modifying the `impl_traits!` macro. If we break `DlogGroup` into two smaller traits and then break up the macro, it would be possible, but that seems like it could be a substantial change. Do you have an suggestions on how to do it?
> 
> Yeah agreed. Splitting the trait might be the way to go to avoid code duplication. Or, calling the same method from different places. I'll take a look and get back about this.

We now have such a clean split, which should let us avoid code duplication: https://github.com/microsoft/Nova/pull/380

srinathsetty:
> > We now have such a clean split, which should let us avoid code duplication: #380
> 
> See my comment on that PR: https://github.com/microsoft/Nova/pull/380/files#r2007610670

Right! We can split impl_traits into impl_traits_dlog, impl_traits_dlogext, etc. Then, the GPU code only has provide an alternate version of impl_traits_dlogext. Isn't it?

src/provider/bn256_grumpkin.rs
jacobtrombetta: Using your work from PR #380, this code section has been updated to only inline code related to the MSM.

srinathsetty:
Thanks for the updates @jacobtrombetta!

It looks like we are not calling into blitzar for small-value MSMs. Is there no API in blitzar at the moment for this?

jacobtrombetta:
> Thanks for the updates @jacobtrombetta!
> 
> It looks like we are not calling into blitzar for small-value MSMs. Is there no API in blitzar at the moment for this?

There is an API. We can certainly add it, but the code is not as simple since the `vartime_multiscalar_mul_small` signature doesn't match the blitzar API. It probably makes sense to add in a follow-up PR.

PR #376, opened 2025-03-13
Enabling Nova to be used in `no_std` environments

milosdjurica:
This PR enables Nova to be used in `no_std` environments. After this PR is merged users of this library will have the option to choose if they want `std` or `no_std` version of Nova. 

## `std` functionalities remain unchanged!

Main points for `no_std` part -> 
1. Rayon parallelization is disabled since parallelization can not be used in `no_std` environments.
2. `halo2curves` dependency can not be used as it is not `no_std` compatible. Instead of `halo2curves`, `pasta_curves` crate is used for Pallas and Vesta curves. Secp, Secq, Bn256 and Grumpkin curves are not available **_for now_** in `no_std` environments. 
3. `bincode` dependency version is changed to 2.0.0-rc.3 as the old one was not compatible for `no_std`. 
4. Using `rand_core` + `rand_chacha` for RNG, as OsRng relies on the operating system and is unavailable in no_std environments.
5. In `no_std` using `OnceCell` from `core::cell::OnceCell` instead of `once_cell` crate .
6. Default features are disabled for most of other dependencies.

As I said, Nova's functionality remains unchanged by default, but this gives users an option to use it in a `no_std` environments too by disabling default features. Also, all tests are passing!
To build the Nova library on `no_std` target, use `cargo build --target thumbv7em-none-eabi --no-default-features`.

I’m more than open to all questions, suggestions, discussions, and corrections! Please let me know what you think! :)


milosdjurica:
> @milosdjurica please read the following Contributor License Agreement(CLA). If you agree with the CLA, please reply with the following information.
> 
> ```
> @microsoft-github-policy-service agree [company="{your company}"]
> ```
> 
> > Options:
> > 
> > * (default - no company specified) I have sole ownership of intellectual property rights to my Submissions and I am not making Submissions in the course of work for my employer.
> > 
> > ```
> > @microsoft-github-policy-service agree
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 
> > * (when company given) I am making Submissions in the course of work for my employer (or my employer has intellectual property rights in my Submissions by contract or applicable law). I have permission from my employer to make Submissions and enter into this Agreement on behalf of my employer. By signing below, the defined term “You” includes me and my employer.
> > 
> > ```
> > @microsoft-github-policy-service agree company="Microsoft"
> > ```
> 
> Contributor License Agreement

@microsoft-github-policy-service agree company="MVP Workshop"

src/gadgets/nonnative/util.rs
```diff
@@ -232,18 +232,29 @@ impl<Scalar: PrimeField> From<AllocatedNum<Scalar>> for Num<Scalar> {
   }
 }
 
-fn write_be<F: PrimeField, W: Write>(f: &F, mut writer: W) -> io::Result<()> {
-  for digit in f.to_repr().as_ref().iter().rev() {
-    writer.write_u8(*digit)?;
-  }
+// fn write_be<F: PrimeField, W: Write>(f: &F, mut writer: W) -> io::Result<()> {
```
srinathsetty, line 235: Please remove commented code.

Cargo.toml
```diff
@@ -63,6 +87,8 @@ name = "ppsnark"
 harness = false
 
 [features]
-default = ["halo2curves/asm"]
+default = ["std"]
```
srinathsetty, line 90: Can we make asm also default?

src/spartan/snark.rs
```diff
@@ -352,15 +377,28 @@ impl<E: Engine, EE: EvaluationEngineTrait<E>> RelaxedR1CSSNARKTrait<E> for Relax
...
+      let T_y = EqPolynomial::evals_from_points(r_y);
 
-      (0..M_vec.len())
+      #[cfg(feature = "std")]
+      let res = (0..M_vec.len())
         .into_par_iter()
         .map(|i| evaluate_with_table(M_vec[i], &T_x, &T_y))
-        .collect()
+        .collect();
+      #[cfg(not(feature = "std"))]
```
srinathsetty, line 395: This will be hard to maintain. Is there a way to use `maybe_rayon` to avoid having to duplicate code?

srinathsetty:
Thanks for submitting this PR! I'll add comments one by one.

As a note: We should also add a CI job to check if the code builds in `no_std` mode.

src/nova/mod.rs
```diff
@@ -742,6 +788,7 @@ where
     );
 
     // create SNARKs proving the knowledge of Wn primary/secondary
+    #[cfg(feature = "std")]
```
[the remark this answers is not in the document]
milosdjurica: I looked into it, but this crate hasn't been maintained for 2+ years and it has only 1 star on GitHub. Even when I tried to use it, it was NOT compiling with `cargo build --target thumbv7em-none-eabi --no-default-feature` command. So, I looked for alternatives and found this crate -> https://crates.io/crates/plonky2_maybe_rayon . I made a changes to use this crate as you can see on the following commit -> https://github.com/microsoft/Nova/pull/376/commits/58b6783ed14ff700b0ba6b56684372e5b9d1e662

src/spartan/snark.rs
milosdjurica: Done! Explained it in the previous comment -> https://github.com/microsoft/Nova/pull/376#discussion_r1995399283

src/nova/mod.rs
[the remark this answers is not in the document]
srinathsetty: This seems like it includes the entire plonky2 as a dependency. I think maybe_rayon is a simple 20-50 line macros. Could we simply include those macros?

Cargo.toml
srinathsetty: The original intention was that the default features give the best performance on x86 (a common target platform). Other platforms could pass --no-default-features.

milosdjurica:
```toml

# If we include asm as default, users who want STD features and are NOT on x86, will have to use it like this ->
nova-snark = { version = "*", default-features = false, features = ["std"] }

# If we leave ASM out by default, users will have to add it manually like this -> 
nova-snark = { version = "*", features = ["asm"] }

```

I personally prefer second approach as it is cleaner, and I believe that majority of users will actually want to use STD features, so I would make `asm` feature optional. Second approach is simpler, cleaner, more "standardized", users wont get bugs if they aren't on `x86_644 aarch` and use Nova with default features. 
After all, even `halo2curves` makes `asm` as optional dependency. 

In the end, it is your decision. Let me know what you prefer, we can do it in both ways

src/digest.rs
```diff
@@ -80,47 +81,38 @@ impl<'a, F: PrimeField, T: Digestible> DigestComputer<'a, F, T> {
...
 
   type E = PallasEngine;
 
   #[derive(Serialize, Deserialize)]
   struct S<E: Engine> {
     i: usize,
-    #[serde(skip, default = "OnceCell::new")]
-    digest: OnceCell<E::Scalar>,
+    #[serde(skip)]
+    digest: Option<E::Scalar>,
```
srinathsetty, line 94: The OnceCell has the property that when it is accessed we compute the digest. Is that same behavior available here?

milosdjurica: This change is taken from your initial no_std branch -> https://github.com/microsoft/Nova/compare/main...no_std#diff-ad10c9494ad2ca56a972214684f5e11ad3c91a75d259f1d8a0ae9d576fc6ec99R96

src/frontend/constraint_system.rs
```diff
@@ -1,8 +1,9 @@
-use std::{io, marker::PhantomData};
-
-use ff::PrimeField;
-
 use super::lc::{Index, LinearCombination, Variable};
+#[cfg(not(feature = "std"))]
+use crate::prelude::*;
+use ff::PrimeField;
+#[cfg(feature = "std")]
+use std::marker::PhantomData;
```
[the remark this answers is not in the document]
milosdjurica: resolved -> https://github.com/microsoft/Nova/pull/376/commits/90679f9d421abe8f8ec85ebac784fc52cb336e82

examples/and.rs
```diff
@@ -1,9 +1,9 @@
 //! This example executes a batch of 64-bit AND operations.
 //! It performs the AND operation by first decomposing the operands into bits and then performing the operation bit-by-bit.
 //! We execute a configurable number of AND operations per step of Nova's recursion.
+use bincode::config::legacy;
```
srinathsetty, line 4: Is there an alternative to this? May be I'm misunderstanding this. Are we at the risk of this getting deprecated soon?

Cargo.toml
srinathsetty: Sounds good!

src/digest.rs
srinathsetty: Right, but that was not meant to be any suggestion. Is there no way to have oncecell?

examples/and.rs
milosdjurica: No, there is no risk of getting deprecated. `legacy()` is just a function to get same encoding configuration that was used in v1.x.x . I used `legacy()` over standard, just in order to make it backward compatible with previous version of Nova. 
Here is the `bincode` code explaining the difference between the `standard()` and `legacy()` functions. 

https://github.com/bincode-org/bincode/blob/508a00bb17fda81da1ad2b3856669c27b7fe5dbd/src/config.rs#L52


```rust
/// The default config for bincode 2.0. By default this will be:
/// - Little endian
/// - Variable int encoding
pub const fn standard() -> Configuration {
    generate()
}

/// Creates the "legacy" default config. This is the default config that was present in bincode 1.0
/// - Little endian
/// - Fixed int length encoding
pub const fn legacy() -> Configuration<LittleEndian, Fixint, NoLimit> {
    generate()
}


```

src/digest.rs
milosdjurica: Fixed -> https://github.com/microsoft/Nova/pull/376/commits/5804c192e2eca87b5236c639891812665a8ae9c8

examples/and.rs
```diff
@@ -286,9 +286,9 @@ fn main() {
     assert!(res.is_ok());
     let compressed_snark = res.unwrap();
 
-    let mut encoder = ZlibEncoder::new(Vec::new(), Compression::default());
-    bincode::serialize_into(&mut encoder, &compressed_snark).unwrap();
-    let compressed_snark_encoded = encoder.finish().unwrap();
+    // let mut encoder = ZlibEncoder::new(Vec::new(), Compression::default());
```
srinathsetty, line 289: Nit: please remove commented line.

src/digest.rs
```diff
@@ -143,19 +155,22 @@ mod tests {
...
-    let good_s = S::<E>::new(42);
-
     // let's set up a struct with a weird digest field to confuse deserializers
     let oc = OnceCell::new();
     oc.set(<E as Engine>::Scalar::ONE).unwrap();
 
-    let bad_s: S<E> = S { i: 42, digest: oc };
+    let mut good_s = S::<E>::new(42);
+    let mut bad_s: S<E> = S { i: 42, digest: oc };
+    // this justifies the adjective "bad"
```
srinathsetty, line 164: Nit: this appears to be a duplicate comment

examples/and.rs
srinathsetty: I'm not sure there's benefit from being backward compatible with a previous version of Nova. If we are updating to bincode 2.0, I think it makes sense to have the bincode 2.0 format.

examples/and.rs
```diff
@@ -286,9 +286,8 @@ fn main() {
     assert!(res.is_ok());
     let compressed_snark = res.unwrap();
 
-    let mut encoder = ZlibEncoder::new(Vec::new(), Compression::default());
```
srinathsetty, line 289: It seems like we are removing flate2 compression. Is there a reason?

src/digest.rs
```diff
@@ -1,29 +1,27 @@
...
+use bincode::config::legacy;
+use core::marker::PhantomData;
 use ff::PrimeField;
 use serde::Serialize;
 use sha3::{Digest, Sha3_256};
-use std::{io, marker::PhantomData};
 
 /// Trait for components with potentially discrete digests to be included in their container's digest.
 pub trait Digestible {
   /// Write the byte representation of Self in a byte buffer
```
srinathsetty, line 12: Nit: the comment could be updated to say "we return a byte vector"?

src/gadgets/nonnative/bignat.rs
```diff
@@ -3,13 +3,18 @@ use super::{
...
 };
 use crate::frontend::{ConstraintSystem, LinearCombination, SynthesisError};
+#[cfg(not(feature = "std"))]
+use crate::prelude::*;
 use ff::PrimeField;
+use libm::log2;
 use num_bigint::BigInt;
 use num_traits::cast::ToPrimitive;
+#[cfg(feature = "std")]
 use std::{
```
srinathsetty, line 13: These imports are present in core. Would changing it to be from core instead of std simply this?

src/gadgets/nonnative/util.rs
```diff
@@ -2,13 +2,13 @@ use super::{BitAccess, OptionExt};
 use crate::frontend::{
   num::AllocatedNum, ConstraintSystem, LinearCombination, SynthesisError, Variable,
 };
-use byteorder::WriteBytesExt;
+#[cfg(not(feature = "std"))]
+use crate::prelude::*;
+// use byteorder::WriteBytesExt;
```
srinathsetty, line 7: Nit: remove commented code

src/provider/hyperkzg.rs
```diff
@@ -1045,11 +1059,13 @@ mod tests {
...
     assert_eq!(post_c_p, post_c_v);
 
-    let proof_bytes = bincode::DefaultOptions::new()
-      .with_big_endian()
-      .with_fixint_encoding()
-      .serialize(&proof)
-      .unwrap();
+    let proof_bytes = bincode::serde::encode_to_vec(&proof, legacy()).unwrap();
+
+    // let proof_bytes = bincode::DefaultOptions::new()
```
srinathsetty, line 1064: Nit: please remove commented code.

src/provider/mod.rs
```diff
@@ -1,12 +1,20 @@
...
 pub mod hyperkzg;
 pub mod ipa_pc;
+#[cfg(feature = "std")]
 pub mod msm;
+#[cfg(not(feature = "std"))]
+pub mod msm_no_std_impl;
+#[cfg(feature = "std")]
 pub mod pasta;
+#[cfg(not(feature = "std"))]
+pub mod pasta_curves;
```
srinathsetty, line 15: Why do we need a separate file?

src/provider/mod.rs
```diff
@@ -1,12 +1,20 @@
...
 
 // public modules to be used as an evaluation engine with Spartan
+#[cfg(feature = "std")]
 pub mod bn256_grumpkin;
 pub mod hyperkzg;
 pub mod ipa_pc;
+#[cfg(feature = "std")]
 pub mod msm;
+#[cfg(not(feature = "std"))]
+pub mod msm_no_std_impl;
```
srinathsetty, line 11: Could we have the no-std impl of msm in the same file?

src/provider/msm_no_std_impl.rs
```diff
@@ -0,0 +1,504 @@
+#![cfg(not(feature = "std"))]
```
srinathsetty, line 1: I'm not sure this code is now necessary. With a recent PR, Nova has its own faster MSM methods. That code should be no-std compatible. Could we just use that code?

src/provider/pasta_curves.rs
```diff
@@ -0,0 +1,176 @@
+#![cfg(not(feature = "std"))]
+//! This module implements the Nova traits for `pallas::Point`, `pallas::Scalar`, `vesta::Point`, `vesta::Scalar`.
+
```
srinathsetty, line 3: We recently dropped support for pasta_curves because it is hard to maintain layers on multiple curve libraries and halo2curves implements pasta_curves. Is there a reason to bring this back?

src/r1cs/mod.rs
```diff
@@ -397,13 +414,24 @@ impl<E: Engine> R1CSShape<E> {
...
+    let Z = (0..self.num_vars + self.num_io + 1)
+      .map(|_| E::Scalar::random(&mut ChaCha20Rng::seed_from_u64(0xDEADBEEF)))
+      .collect::<Vec<E::Scalar>>();
 
+    #[cfg(feature = "std")]
     let r_W = E::Scalar::random(&mut OsRng);
+    #[cfg(feature = "std")]
     let r_E = E::Scalar::random(&mut OsRng);
+    #[cfg(not(feature = "std"))]
+    let r_W = E::Scalar::random(&mut ChaCha20Rng::seed_from_u64(0xDEADBEEF));
```
srinathsetty, line 432: This is not safe. We need to sample random numbers to provide zero-knowledge. Please change these everywhere to call into a secure random number generator.

src/spartan/mod.rs
```diff
@@ -275,5 +315,21 @@ fn compute_eval_table_sparse<E: Engine>(
     },
   );
 
+  // Sequential execution instead of rayon::join
+  #[cfg(not(feature = "std"))]
```
srinathsetty, line 319: Is there a reason to do this here than in the underlying code `inner`?

srinathsetty:
Hi @milosdjurica thanks again for your work on this! I've left some comments. The most important one is related to random number generator used. It seems like the code hardcodes randomness instead of sampling from a secure rng. Is there no random number generator in no-std environment? If yes, then I'm not sure the prover and the setup can be safely implemented in such environments.

src/digest.rs
```diff
@@ -45,15 +43,15 @@ impl<'a, F: PrimeField, T: Digestible> DigestComputer<'a, F, T> {
     });
 
     // turn the bit vector into a scalar
-    let mut digest = F::ZERO;
```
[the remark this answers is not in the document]
milosdjurica: Change from here -> https://github.com/microsoft/Nova/compare/main...no_std#diff-ad10c9494ad2ca56a972214684f5e11ad3c91a75d259f1d8a0ae9d576fc6ec99R49

Anyway, I have just reverted the change now

src/gadgets/nonnative/util.rs
```diff
@@ -232,18 +232,21 @@ impl<Scalar: PrimeField> From<AllocatedNum<Scalar>> for Num<Scalar> {
...
   }
-
   Ok(())
 }
 
 /// Convert a field element to a natural number
 pub fn f_to_nat<Scalar: PrimeField>(f: &Scalar) -> BigInt {
-  let mut s = Vec::new();
-  write_be(f, &mut s).unwrap(); // f.to_repr().write_be(&mut s).unwrap();
+  let mut buffer = [0u8; 256]; // Fixed-size buffer for the serialized representation of the field element
```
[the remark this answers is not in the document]
milosdjurica: https://github.com/microsoft/Nova/compare/main...no_std#diff-64fe0ff2f8bf7ebf971a810101c2090559d4257be9de85ac552eac3da37cdf47R245

Not completely sure, it was your change from experimental no_std branch

src/frontend/lc.rs
```diff
@@ -1,7 +1,9 @@
-use std::ops::{Add, Sub};
-
+#[cfg(not(feature = "std"))]
+use crate::prelude::*;
 use ff::PrimeField;
 use serde::{Deserialize, Serialize};
+#[cfg(feature = "std")]
+use std::ops::{Add, Sub};
```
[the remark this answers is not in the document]
milosdjurica: https://github.com/microsoft/Nova/pull/376/commits/5b26ac4969dabc1b6b815f56b0f4fc588b0c63e2

src/gadgets/nonnative/bignat.rs
milosdjurica: https://github.com/microsoft/Nova/pull/376/commits/5b26ac4969dabc1b6b815f56b0f4fc588b0c63e2

src/provider/mod.rs
milosdjurica: Fixed -> https://github.com/microsoft/Nova/pull/376/commits/6d4c9596334992940c2198b42dfe650985dd7f20

src/provider/msm_no_std_impl.rs
milosdjurica: Fixed -> https://github.com/microsoft/Nova/pull/376/commits/6d4c9596334992940c2198b42dfe650985dd7f20

src/provider/mod.rs
```diff
@@ -15,33 +23,44 @@ pub(crate) mod pedersen;
...
-    secp_secq::{secp256k1, secq256k1},
   },
   traits::Engine,
 };
+#[cfg(feature = "std")]
 pub use ptau::{check_sanity_of_ptau_file, read_ptau, write_ptau};
 use serde::{Deserialize, Serialize};
 
 /// An implementation of Nova traits with HyperKZG over the BN256 curve
+#[cfg(feature = "std")]
```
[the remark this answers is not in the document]
milosdjurica: I agree, it is not ideal scenario. The current problem is that `halo2curves` are not `no_std` compatible. I started working on converting `halo2curves` crate to be `no_std` compatible, then we will be able to use `halo2curves` everywhere (in std & no_std environment). 

You can see the changes here -> https://github.com/privacy-scaling-explorations/halo2curves/compare/main...milosdjurica:halo2curves:no_std

Any help with those changes is more than welcome!

src/provider/pasta_curves.rs
milosdjurica: Responded here -> https://github.com/microsoft/Nova/pull/376#discussion_r2001453868

src/provider/hyperkzg.rs
```diff
@@ -454,7 +462,12 @@ where
 
   fn setup(label: &'static [u8], n: usize) -> Self::CommitmentKey {
     // NOTE: this is for testing purposes and should not be used in production
-    Self::CommitmentKey::setup_from_rng(label, n, OsRng)
+    #[cfg(feature = "std")]
+    let res = Self::CommitmentKey::setup_from_rng(label, n, OsRng);
+    #[cfg(not(feature = "std"))]
+    let res = Self::CommitmentKey::setup_from_rng(label, n, ChaCha20Rng::seed_from_u64(0xDEADBEEF));
```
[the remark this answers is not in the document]
milosdjurica: Hey, I think you are right and this probably is not safe. I will look into this and try to find an alternative in case that there is no secure way to use this. If you have any suggestion please let me know :)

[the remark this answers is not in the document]
srinathsetty: I don't know what the standard practice is to get cryptographically strong random numbers in no_std mode. It would be worth investigating.

src/provider/mod.rs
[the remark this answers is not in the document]
srinathsetty: I think there may be a way to pass pasta_curves to Nova without needing Nova to maintain that code. For example, we could have the layer above Nova implement Nova's traits for pasta_curves types right?

examples/and.rs
milosdjurica: Yes, due to the changes in `bincode` . If we use `bincode::serde::::encode_to_vec()` , that function accepts only 2 arguments -> what we are encoding (compressed_snark in this case), and bincode configuration. 

If we use `bincode::serde::encode_into_writter()` , this function accepts 3 arguments -> what we are encoding, "writer", and bincode configuration. However, writer has to implement `bincode::Writer` trait, and this requirement is not satisfied.

milosdjurica:
Hi @srinathsetty , here is the update. 

### Halo2curves

Regarding halo2curves, I have opened PR to make it `no_std`, now waiting for it to get merged. After it gets merged, we will be able to use `halo2curves` in both -> `no_std` and `std` environment, and there won't be need for `pasta_curves` crate. Link to PR -> https://github.com/privacy-scaling-explorations/halo2curves/pull/192

There is one more thing regarding this. Nova is using `halo2curves` with `derive_serde` feature, and this includes one more crate which is `serde_arrays` and it is not no_std compatible, so I have opened a PR there too -> https://github.com/Kromey/serde_arrays/pull/36


### RNG

Regarding RNG in no_std, seems like there is no universal and simple way to get secure random number (at least I am not aware of it right now). ChaCha is creating a PRNG from seed, but for the same seed, output will always be the same and predictable, which is not secure. The solution is to get a random number in no_std environment is to provide a ChaCha with secure entropy source, but this depends on a specific targets.

Under the hood, `rand` and `rand_core` libraries use `getrandom` crate to get random data from system resources. In order to get this to work with `thumbv7em-none-eabi` target, we would need to create our own backend that will retrieve random data from system.

### Possible solution for RNG 

What we can do is to hide functionality that requires RNG behind STD feature flag. This would mean that STD users can do same things as they always could. In no_std mode users would only be able to verify the proofs, but not create them. This is still improvement, as in earlier versions, no_std users could not interact with library in any way.

- std -> can prove and verify (stays the same like it has always been and uses safe OsRng)
- no-std -> can only verify

Please let me know what you think. Thanks!

srinathsetty:
> * std -> can prove and verify (stays the same like it has always been and uses safe OsRng)
> * no-std -> can only verify

It is dangerous to provide APIs where someone can produce a proof in no_std mode that are easily faked.

milosdjurica:
> Update: It is dangerous to provide APIs where someone can produce a proof in no_std mode that are easily distinguishable. In particular, if the prover uses a seed that's known or can be guessed, it's easy to break zero-knowledge. We cannot hope that no one runs the prover in no_std mode.


I completely agree with this, but this is not what I was suggesting. I will give a minimal example to explain what I was trying to say.


```rust
struct CompressedSNARK {
  // Fields...
}

impl CompressedSNARK {
  #[cfg(feature = "std")]
  pub fn prove() {
    // This method uses RNG and it will only be available in STD mode. We will keep using OsRng as before.
    // But, users WILL NOT BE ABLE to `prove()` in no_std environment due to the feature flag above the method -> `#[cfg(feature = "std")]`.
    // This means that if users want to call this method, they must be in STD mode, and they will use safe OsRng for random number.
  }
  pub fn verify() {
    // This method DOES NOT use RNG, so it is completely safe to let users call this function in both, std and no_std mode.
  }
}


`````

Users in STD environment can call both `prove` and `verify` methods, but users in NO_STD environment can ONLY call `verify` method. 

### Long story short ->
- NO_STD users will only be able to use safe parts of API (parts that do not use RNG). 
- STD part will stay completely same and completely safe as before. It will use safe `OsRng` for creating random numbers as it was using it before.

-------- 

> If your goal is to only run the verifier in no_std mode, then why not simply write a stand-alone program that deserializes Nova's bincode-encoded proofs and verify them? The verification algorithm can be easily pieced together using code provided in this library.

Regarding this : In order to deserialize proof and verify it, you need to specify in which struct you want to deserialize into. That means you need to import a struct from Nova, and that means Nova has to be dependency in your project. If Nova is not no_std compatible, then the program calling it also can not be no_std compatible.

---

Let me know what you think about this. I hope I explained it well now, and that it makes more sense. If you agree with this, I will make another commit that will remove using unsafe RNG in `no_std` mode and only enable using safe RNG in `std` mode

---


P.S. -> Latest commit https://github.com/microsoft/Nova/pull/376/commits/b45a5db9e79169917d10ae4b2e77b20df1139c15 removes `pasta_curves` completely as dependency and enables using `halo2curves` in both STD and NO_STD mode. 
I am waiting for the PR to be accepted in `halo2curves` and then I will change Cargo.toml to point to their repo instead of my fork.

File: src/provider/mod.rs (hunk 24 above)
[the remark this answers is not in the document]
Comment (milosdjurica) [reply src/provider/mod.rs]: Latest commit https://github.com/microsoft/Nova/pull/376/commits/b45a5db9e79169917d10ae4b2e77b20df1139c15 resolves this by removing pasta_curves completely, and uses halo2curves for STD and NO_STD. This will make code much more simpler to maintain going forward. I am waiting for PR in halo2curves to be accepted and then I will change halo2curves import in Cargo.toml -> https://github.com/microsoft/Nova/commit/b45a5db9e79169917d10ae4b2e77b20df1139c15#diff-2e9d962a08321605940b5a657135052fbcef87b5e360662bb527c96d9a615542R46

File: src/provider/pasta_curves.rs (hunk 18 above)
Comment (milosdjurica) [reply src/provider/pasta_curves.rs]: I have made changes in the latest commit to resolve this, now I am just waiting for the PR in halo2curves to be accepted. Changes -> https://github.com/microsoft/Nova/commit/b45a5db9e79169917d10ae4b2e77b20df1139c15

**Comment by milosdjurica** [comment]
Hi @srinathsetty ,

Here is commit regarding RNG -> https://github.com/microsoft/Nova/pull/376/commits/402073b5bc7625a8ca5f77626d76a8fd1dc0ea68

I decided to take a bit different approach from what I have explained earlier, as previous approach would bloat the whole codebase with `#[cfg(feature = "std")]` flags. 
What I chose to do was to introduce a new error - `NovaError::UnsafeRandomNumber` , and throw that error when user tries to call RNG method and `std` feature is off.

This keeps the codebase clean, safe and easy to maintain. 
Let me know what you think of this. Thank you!

**Comment by srinathsetty** [comment]
> Regarding this : In order to deserialize proof and verify it, you need to specify in which struct you want to deserialize into. That means you need to import a struct from Nova, and that means Nova has to be dependency in your project. If Nova is not no_std compatible, then the program calling it also can not be no_std compatible.

Diving a bit into this. I'm not sure it is necessary to have Nova as a dependency at all. There should be a way to build a stand-alone `nova-verifier` that is pinned to a particular nova prover version that can deserialize proofs produced by the prover into types owned by this project and verify them. We did this in a completely different language (Solidity), see e.g., https://eprint.iacr.org/2024/2099, so it should be possible to do the same in Rust.

**Comment by milosdjurica** [comment]
> Diving a bit into this. I'm not sure it is necessary to have Nova as a dependency at all. There should be a way to build a stand-alone nova-verifier that is pinned to a particular nova prover version that can deserialize proofs produced by the prover into types owned by this project and verify them. We did this in a completely different language (Solidity), see e.g., https://eprint.iacr.org/2024/2099, so it should be possible to do the same in Rust.


Hi @srinath, Thanks for this, I will have a look.

Do you maybe have any other suggestions for other changes made in this PR? 

P.S.
I did some tests about faster MSM https://github.com/microsoft/Nova/issues/372 and i found out that halo2curves MSM implementation is almost 2x faster than the current one. However, this would require more testing as I only tested for verifications of Pallas-Vesta SNARKs with spark compression, without any additional features turned on. Values I got are 412ms for current MSM implementation in this library, and 230ms for halo2curves MSM implementation.

I tested in both STD and NO_STD environments and in both cases it was faster. Might be worth exploring this deeper. You can see changes here ->  https://github.com/MVPWorkshop/Nova/commit/6c30073c098425fd55c5541c321ca03e36b88f27 . Note: this is not included in this current PR.

`msm_fast.rs` is halo2curves implementation from https://github.com/privacy-scaling-explorations/halo2curves/blob/main/src/msm.rs with some small changes in order to make it compile on `no_std`.

**Comment by srinathsetty** [comment]
> > Diving a bit into this. I'm not sure it is necessary to have Nova as a dependency at all. There should be a way to build a stand-alone nova-verifier that is pinned to a particular nova prover version that can deserialize proofs produced by the prover into types owned by this project and verify them. We did this in a completely different language (Solidity), see e.g., https://eprint.iacr.org/2024/2099, so it should be possible to do the same in Rust.
> 
> Hi @srinathsetty , Thanks for this, I will take a look.
> 
> Do you maybe have any other suggestions for other changes made in this PR?
> 
> P.S. I did some tests about faster MSM #372 and i found out that halo2curves MSM implementation is almost 2x faster than the current one. However, this would require more testing as I only tested for verifications of Pallas-Vesta SNARKs with spark compression, without any additional features turned on. Values I got are 412ms for current MSM implementation in this library, and 230ms for halo2curves MSM implementation.
> 
> I tested in both STD and NO_STD environments and in both cases it was faster. Might be worth exploring this deeper. You can see changes here -> [MVPWorkshop@6c30073](https://github.com/MVPWorkshop/Nova/commit/6c30073c098425fd55c5541c321ca03e36b88f27) . Note: this is not included in this current PR.
> 
> `msm_fast.rs` is halo2curves implementation from https://github.com/privacy-scaling-explorations/halo2curves/blob/main/src/msm.rs with some small changes in order to make it compile on `no_std`.

This is unexpected. Would you be able to run the commitment benchmark on your machine and report results here. The command is: cargo bench --bench=commit

**Comment by milosdjurica** [comment]
You are right regarding the benchmarks. I have ran benchmarks and it says that halo2curves is slower. However, verification says different.

Could you update your test `src/nova/mod.rs::tests::test_ivc_nontrivial_with_spark_compression_with<...>(...)` like this -> 

```diff
  fn test_ivc_nontrivial_with_spark_compression_with<E1, E2, EE1, EE2>()
  where
    E1: Engine<Base = <E2 as Engine>::Scalar>,
    E2: Engine<Base = <E1 as Engine>::Scalar>,
    EE1: EvaluationEngineTrait<E1>,
    EE2: EvaluationEngineTrait<E2>,
  {
    // ...
    // ...

+   let start = std::time::Instant::now();
    // verify the compressed SNARK
    let res = compressed_snark.verify(&vk, num_steps, &[<E1 as Engine>::Scalar::ZERO]);
+   let duration = start.elapsed();
+   println!("Duration is -> {:?}", duration);



```

and run test with the following command ->  `cargo test --release -- --nocapture test_ivc_nontrivial_with_` 
`-- --nocapture` allows us to see the prints. 
Do this with the current msm implementation on main and then compare it with halo2curves msm implmentation. https://github.com/microsoft/Nova/compare/main...MVPWorkshop:Nova:proof_of_faster_msm Here you can see the changes needed in order to use halo2curves msm implementation.

Those are my results -> 

ON MAIN -> 
```
running 1 test
Duration is -> 473.631666ms
Duration is -> 274.569834ms
test nova::tests::test_ivc_nontrivial_with_spark_compression has been running for over 60 seconds
Duration is -> 545.656791ms
test nova::tests::test_ivc_nontrivial_with_spark_compression ... ok
```

USING HALO2CURVES MSM ->
```
running 1 test
Duration is -> 223.310875ms
Duration is -> 118.1095ms
Duration is -> 225.901208ms
test nova::tests::test_ivc_nontrivial_with_spark_compression ... ok
```


It is faster in all 3 cases, and it should also be faster not only for spark compression but also for normal compression in ->  `test_ivc_nontrivial_with_compression()`

srinathsetty:
> You are right regarding the benchmarks. I have ran benchmarks and it says that halo2curves is slower. However, verification test says different.
> 
> Could you update your test `src/nova/mod.rs::tests::test_ivc_nontrivial_with_spark_compression_with<...>(...)` like this ->
> 
> ```diff
>   fn test_ivc_nontrivial_with_spark_compression_with<E1, E2, EE1, EE2>()
>   where
>     E1: Engine<Base = <E2 as Engine>::Scalar>,
>     E2: Engine<Base = <E1 as Engine>::Scalar>,
>     EE1: EvaluationEngineTrait<E1>,
>     EE2: EvaluationEngineTrait<E2>,
>   {
>     // ...
>     // ...
> 
> +   let start = std::time::Instant::now();
>     // verify the compressed SNARK
>     let res = compressed_snark.verify(&vk, num_steps, &[<E1 as Engine>::Scalar::ZERO]);
> +   let duration = start.elapsed();
> +   println!("Duration is -> {:?}", duration);
> ```
> 
> and run test with the following command -> `cargo test --release -- --nocapture test_ivc_nontrivial_with_` . Flag `-- --nocapture` allows us to see the prints. Do this with the current msm implementation on main and then compare it with halo2curves msm implmentation. [main...MVPWorkshop:Nova:proof_of_faster_msm](https://github.com/microsoft/Nova/compare/main...MVPWorkshop:Nova:proof_of_faster_msm) Here you can see the changes needed in order to use halo2curves msm implementation.
> 
> Those are my results ->
> 
> ON MAIN ->
> 
> ```
> running 1 test
> Duration is -> 473.631666ms
> Duration is -> 274.569834ms
> test nova::tests::test_ivc_nontrivial_with_spark_compression has been running for over 60 seconds
> Duration is -> 545.656791ms
> test nova::tests::test_ivc_nontrivial_with_spark_compression ... ok
> ```
> 
> USING HALO2CURVES MSM ->
> 
> ```
> running 1 test
> Duration is -> 223.310875ms
> Duration is -> 118.1095ms
> Duration is -> 225.901208ms
> test nova::tests::test_ivc_nontrivial_with_spark_compression ... ok
> ```
> 
> It is faster in all 3 cases, and it should also be faster not only for spark compression but also for normal compression in -> `test_ivc_nontrivial_with_compression()`

It is not clear why the microbenchmark will show halo2curves as slow but faster when used elsewhere. Thanks, will investigate!

PR #382, opened 2025-03-26
perf: remove 16 bytes from hyperkzg proof by transposing v

JayWhite2357:
When using serde, the `v` field in the hyperkzg proof has its length encoded three times.
Transposing `v` to use `Vec<[E::Scalar; 3]>` instead of `[Vec<E::Scalar>; 3]` encodes it only once, resulting in a savings of 16 bytes when using bincode-fixint.

### DIFF
--- a/src/provider/hyperkzg.rs
+++ b/src/provider/hyperkzg.rs
@@ -581,13 +581,17 @@ where
 {
   com: Vec<G1Affine<E>>,
   w: [G1Affine<E>; 3],
-  v: [Vec<E::Scalar>; 3],
+  v: Vec<[E::Scalar; 3]>,
 }
 
 impl<E: Engine> EvaluationArgument<E>
 where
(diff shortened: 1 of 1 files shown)


src/provider/hyperkzg.rs
```diff
@@ -581,13 +581,17 @@ where
...
-  v: [Vec<E::Scalar>; 3],
+  v: Vec<[E::Scalar; 3]>,
 }
 
 impl<E: Engine> EvaluationArgument<E>
 where
   E::GE: PairingGroup,
 {
+  /// Create a new evaluation argument
+  pub fn new(com: Vec<G1Affine<E>>, w: [G1Affine<E>; 3], v: Vec<[E::Scalar; 3]>) -> Self {
```
srinathsetty, line 592: Nit: is this method necessary?

PR #403, opened 2025-08-03
feat: bump bincode to 2

iajoiner:
Bump bincode to 2 since 1.3.3 is 4 years old.


iajoiner:

PR #404, opened 2025-08-03
feat: bump halo2curves to 0.9.0

iajoiner:
halo2curves 0.9.0 can be easily used in WASM so it is a good idea to bump it.


srinathsetty:
@iajoiner Thanks for the PR! Can you please fix the wasm build issue?

PR #405, opened 2025-08-04
Mercury PCS

Sun-Jc:
[Mercury](https://eprint.iacr.org/2025/385.pdf) is implemented as an evaluation engine.

### DIFF
--- a/src/provider/mercury.rs
+++ b/src/provider/mercury.rs
@@ -5,7 +5,9 @@
 //! The opening proof size is constant.
 //! The verification consists of O(log n) field operations and O(1) scalar multiplications, along with 2 pairings.
 //!
-//! Mercury and HyperKZG shares the same commitment key and engine.
+//! The batch KZG evaluation tech is from BDFG20<https://eprint.iacr.org/2020/081.pdf>
+//!
+//! Mercury and HyperKZG shares the same types of commitment key and engine.
 //!
 //! Samaritan presents a similar construction and achieves the same performance, see <https://eprint.iacr.org/2025/419.pdf>.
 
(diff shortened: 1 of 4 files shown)


src/provider/mercury.rs
```diff
@@ -0,0 +1,1427 @@
...
+        comm_d: arg.comm_d,
+        comm_w: arg.comm_w,
+        comm_w_prime: arg.comm_w_prime,
+      },
+      g1,
+      transcript,
+    )?;
+
+    // Check Pairing of 3. and 4.
+    let d = E::Scalar::random(rand_core::OsRng);
```
srinathsetty: I think we want to obtain this challenge from the transcript. The main reason is otherwise an on-chain verifier cannot do the same.

srinathsetty: A side effect of obtaining this challenge from the transcript is that the prover has to squeeze a challenge as well to keep the transcript in the same state.

Sun-Jc:
> Could we also introduce Engines so we can use Mercury in some of the end-to-end test cases of IVC and ppsnark?

Hi @srinathsetty , do you mind describing the question in another way? 

`traits::Engine` and `*::EvaluationEngine` are combined as follows:
(Engine does not rely on EvaluationEngine)

```rust
type E1 = Bn256EngineKZG;
type E2 = GrumpkinEngine;
type EE = nova_snark::provider::hyperkzg::EvaluationEngine<E1>;
type S = nova_snark::spartan::ppsnark::RelaxedR1CSSNARK<E1, EE>;
```

srinathsetty:
Ah good catch!

src/provider/mercury.rs
Sun-Jc: Sure. Done in [Prover](https://github.com/microsoft/Nova/blob/250c83005fe5be6571d4019039275260e516d862/src/provider/mercury.rs#L1131) and [Verifier](https://github.com/microsoft/Nova/blob/250c83005fe5be6571d4019039275260e516d862/src/provider/mercury.rs#L1351)

src/provider/mercury.rs
```diff
@@ -0,0 +1,1440 @@
...
+      if log_n % 2 == 1 {
+        point.insert(0, E::Scalar::ZERO);
+      }
+
+      point
+    };
+
+    let log_n = point.len();
+
+    let u_row = point.split_at(log_n / 2).0.to_vec();
```
srinathsetty, line 1227: There is a requirement here that log_n is at least 2 right? If so, it will be good to check early on to ensure we don't underflow?

Sun-Jc: There is an early check: https://github.com/microsoft/Nova/blob/c710be7eddf55668c60942f312043ab18bac9e9d/src/provider/mercury.rs#L809

PR #410, opened 2025-08-23
fix(bn256): correct G2 group_params() to use proper curve coefficients

MozirDmitriy:
Fix error in G2::group_params() implementation for BN256 curve

The previous implementation incorrectly used G1 curve parameters (bn256::Point::a() 
and bn256::Point::b()) for the G2 group, which is mathematically incorrect since G2 
operates over a quadratic extension field Fq2, not the base field Fq.

Changes:
- Replace G1 parameters with correct constant terms for G2 curve coefficients
- Set A = 0 and B = 3 (constant terms from the Fq2 representation)
- Add Field trait import to support ZERO and from() methods
- Add detailed comments explaining the quadratic extension field constraint

This fix ensures mathematical correctness while maintaining API compatibility.
The G2 group is primarily used for pairing operations in HyperKZG commitment 
scheme, where these parameters provide the correct curve equation coefficients.


srinathsetty:
Thanks for the PR, @MozirDmitriy! Could you please fix the fmt issue?

PR #416, opened 2025-10-13
chore: fix accumulation logic for `PrimeField` vectors

mdqst:
updated the accumulation from using `.sum()` to an explicit `.fold(F::zero(), |acc, x| acc + x)`.
this ensures correct behavior for all `F: PrimeField` types and resolves the issue with summing field elements.

### DIFF
--- a/src/r1cs/sparse.rs
+++ b/src/r1cs/sparse.rs
@@ -91,7 +91,7 @@ impl<F: PrimeField> SparseMatrix<F> {
         self
           .get_row_unchecked(ptrs.try_into().unwrap())
           .map(|(val, col_idx)| *val * vector[*col_idx])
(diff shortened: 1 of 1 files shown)


mdqst:
[srinathsetty](https://github.com/srinathsetty), `sum()` can be a bit tricky with field types - it goes through the `Sum` trait, which doesn’t always behave as expected for `PrimeField` elements (especially with refs and generic bounds). Using an explicit `fold(F::zero(), |acc, x| acc + x)` makes the accumulation unambiguous and guarantees it stays within the field operations. just safer and clearer

mdqst:
[srinathsetty](https://github.com/srinathsetty), lets merge it

mdqst:
[srinathsetty](https://github.com/srinathsetty) lets merge it

PR #418, opened 2025-11-07
optimized sumcheck prover

Sun-Jc:
Optimized sum check prover to save field operations: During iteration over MLP eval points, changed from eval at X=0, X=2, X=3 to evaluating at X=0 and X=-1 and computing the highest order coefficient (x^3’s coefficient for `prove_cubic` and x^2’s coefficient for `prove_quad`)

### DIFF
--- a/src/spartan/polys/univariate.rs
+++ b/src/spartan/polys/univariate.rs
@@ -34,25 +34,18 @@ impl<Scalar: PrimeField> UniPoly<Scalar> {
     }
   }
 
-  pub fn from_evals(evals: &[Scalar]) -> Self {
-    let n = evals.len();
-    let xs: Vec<Scalar> = (0..n).map(|x| Scalar::from(x as u64)).collect();
-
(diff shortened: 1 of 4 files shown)


src/spartan/polys/univariate.rs
```diff
@@ -22,25 +22,30 @@ pub struct CompressedUniPoly<Scalar: PrimeField> {
...
-      row.push(Scalar::ONE);
-      row.push(x);
-      for j in 2..n {
-        row.push(row[j - 1] * x);
-      }
-      row.push(evals[i]);
-      matrix.push(row);
+  // a x^2 + b x + c
+  // evals: [c, a + b + c, a]
+  pub fn from_evals_deg2(evals: &[Scalar]) -> Self {
```
srinathsetty, line 27: Doesn't Gaussian elimination do this automatically?

Sun-Jc: evals[2] is a, not f(2) = 4a + 2b + c. So original Gaussian elimination does not work for this eval list, I think.

PR #423, opened 2025-12-23
Merge file io methods of `CommitmentKey` into `CommitmentEngineTrait`

TheLitFire:


src/provider/pedersen.rs
```diff
@@ -188,17 +189,32 @@ pub struct CommitmentEngine<E: Engine> {
...
     writer.write_all(&KEY_FILE_HEAD)?;
     let mut points = Vec::with_capacity(self.ck.len() + 1);
     points.push(self.h);
     points.extend(self.ck.iter().cloned());
     write_points(writer, points)
   }
+
+  fn check_sanity_of_key_file(
+    path: impl AsRef<std::path::Path>,
+    _num_g1: usize,
```
srinathsetty: if these parameters are not used, why not cut them?

src/traits/commitment.rs
```diff
@@ -48,11 +49,34 @@ pub trait Len {
...
+/// A trait for saving key to a writer and checking sanity of key file
+pub trait CommitmentKeyFileTrait {
+  /// Saves the key to the provided writer.
+  fn save_to(
+    &self,
+    writer: &mut (impl std::io::Write + std::io::Seek),
+  ) -> Result<(), PtauFileError>;
+
+  /// Checks the sanity of the key file at the provided path.
+  fn check_sanity_of_key_file(
```
srinathsetty: Is this method only needed for testing purposes?

TheLitFire:
This PR moves `save_to` method of HyperKZG and Pedersen that saves commitment key to a file into `CommitmentEngineTrait`, and rename it to `save_setup` , coupled with the existing `load_setup`. With this PR, outer implementers who only have access to `Engine` or `CommitmentEngine` will be able to save&load keys, too.

This PR also gates related tests behind `io` feature.

srinathsetty (approved):

PR #490, opened 2026-04-02
Add ROMode for configurable Poseidon sponge width

srinathsetty:
Add an `ROMode` enum (`Wide` / `Narrow`) to `ROTrait` and `ROCircuitTrait` with `new_with_mode()` constructors, enabling callers to select a reduced-width Poseidon sponge without introducing new type parameters.

## Exisiting code

`ROMode` defaults to `Wide`, `new()` is unchanged, and `PoseidonConstantsCircuit`/`PoseidonRO`/`PoseidonROCircuit` retain their existing type signatures. Callers opt in to narrow mode via `new_with_mode(constants, ROMode::Narrow)`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "nova-snark"
-version = "0.70.0"
+version = "0.71.0"
 authors = ["Srinath Setty <srinath@microsoft.com>"]
 edition = "2021"
 description = "High-speed recursive arguments from folding schemes"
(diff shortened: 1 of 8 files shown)


iontzialla:
Is this fully backwards compatible? Seems like it changes the serialization of PoseidonConstantsCircuit which seems to be part of `PublicParams`. So if any user has previously serialized public params they won't be able to parse it after the updates, right?

srinathsetty:
> Is this fully backwards compatible? Seems like it changes the serialization of PoseidonConstantsCircuit which seems to be part of `PublicParams`. So if any user has previously serialized public params they won't be able to parse it after the updates, right?

Good point! Fixed PR description.

iontzialla (approved):

PR #492, opened 2026-04-03
feat: structural zero wire in R1CS layout (z = [w, 1, 0, x])

sai-deng:
Change the R1CS vector layout from `z = [w, 1, x]` to `z = [w, 1, 0, x]`,
making zero a structural part of the vector like the existing one-wire.
The ONE wire stays at its historical position `num_vars` (unchanged).

- `CS::one()` returns `Input(0)` (unchanged), `CS::zero()` returns `Input(1)` (new)
- All CS defaults initialize with 2 built-in inputs (ONE, ZERO)
- Frontend bridge skips 2 built-in inputs instead of 1
- R1CS core: z-vector length is `num_vars + 2 + num_io`, column validation adjusted
- Spartan verifier MLE evaluates `(u, 0, X)` instead of `(u, X)`
- Add `AllocatedNum::zero()` wrapping `CS::zero()` (zero constraints)
- Remove `alloc_zero()` (was 1 aux + 1 constraint per call, 8 call sites)
- Update all hand-built test R1CS, constraint counts, PP digests

### DIFF
--- a/src/r1cs/mod.rs
+++ b/src/r1cs/mod.rs
@@ -156,7 +156,7 @@ impl<E: Engine> R1CSShape<E> {
      -> Result<Vec<()>, NovaError> {
       M.iter()
         .map(|(row, col, _val)| {
-          if row >= num_cons || col > num_io + num_vars {
+          if row >= num_cons || col > num_io + num_vars + 1 {
             Err(NovaError::InvalidIndex)
           } else {
             Ok(())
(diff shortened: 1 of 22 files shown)


src/r1cs/mod.rs
```diff
@@ -742,16 +753,19 @@ impl<E: Engine> R1CSShape<E> {
...
   ) -> Result<(RelaxedR1CSInstance<E>, RelaxedR1CSWitness<E>), NovaError> {
-    // sample Z = (W, u, X)
-    let Z = (0..self.num_vars + self.num_io + 1)
+    // sample Z = (W, 0, u, X)
+    let mut Z = (0..self.num_vars + self.num_io + 2)
       .into_par_iter()
       .map(|_| E::Scalar::random(&mut OsRng))
       .collect::<Vec<E::Scalar>>();
 
+    // The zero wire must be zero
```
srinathsetty: Is this necessary?

sai-deng: E is computed from Z here, but commit_T and is_sat_relaxed hardcode 0 at this position. A non-zero value would make E inconsistent with the folded equation.

sai-deng: Reordered the layout from [W, 0, 1, X] to [W, 1, 0, X] so the ONE wire stays at its historical position num_vars. This avoids changing every place that extracts u = Z[num_vars] and keeps the Spartan MLE split untouched. The zero wire is now at num_vars + 1 instead. Force-pushed with the updated commit.

PR #495, opened 2026-05-18
Updates gadgets to make namespace pushing more consistent

MercysJest:
- Adds a namespace push when calling `field_into_allocated_bits_le`
- Remove unnecessary helper function `triop` to flatten function call stack.


src/frontend/gadgets/uint32.rs
```diff
@@ -151,10 +149,10 @@ impl UInt32 {
     })
   }
 
-  /// Compute the `maj` value (a and b) xor (a and c) xor (b and c)
```
srinathsetty, line 154: Why do we need to change maj to ch?

srinathsetty (approved):

PR #497, opened 2026-06-11
Harden deserialization paths

srinathsetty:


src/provider/bn256_grumpkin.rs
```diff
@@ -204,10 +206,143 @@ impl crate::traits::evm_serde::CustomSerdeTrait for G2Affine {
...
+    }
+  }
+}
+
+#[cfg(all(test, feature = "evm"))]
+mod evm_serde_tests {
+  use super::bn256;
+  use crate::traits::evm_serde::EvmCompatSerde;
+  use halo2curves::group::{cofactor::CofactorCurveAffine, Curve};
+  use serde::{Deserialize, Serialize};
```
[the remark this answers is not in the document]
srinathsetty: The Copilot review comment is incorrect. CofactorCurveAffine is used in three places (identity() at lines 274 & 315, to_curve() at line 351).

src/provider/ptau.rs
```diff
@@ -409,3 +428,57 @@ where
...
+  use halo2curves::group::Curve;
+  use std::io::Cursor;
+
+  // An on-curve G2 point that is NOT in the prime-order subgroup. bn256 G2 has
+  // a large cofactor, so solving the curve equation for an arbitrary x almost
+  // always yields such a point.
+  fn non_subgroup_g2() -> G2Affine {
+    use ff::Field;
+    use halo2curves::bn256::Fq2;
+    use halo2curves::group::cofactor::{CofactorCurveAffine, CofactorGroup};
```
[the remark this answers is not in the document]
srinathsetty: CofactorCurveAffine provides the to_curve() method called at line 465. Removing it breaks compilation.

src/provider/traits.rs
```diff
@@ -50,6 +50,7 @@ pub trait DlogGroup:
     + for<'de> Deserialize<'de>
     + TranscriptReprTrait<Self>
     + CurveAffine
+    + CofactorCurveAffine
     + SerdeObject
```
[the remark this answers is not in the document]
srinathsetty: This is necessary

src/provider/bn256_grumpkin.rs
abhiramkothapalli, line 1: Are other curves also updated to add these checks? If not we should be clear in the documentation that this hardening is only for Grumpkin.

src/provider/hyperkzg.rs
```diff
@@ -50,6 +50,28 @@ type G2Affine<E> = <<<E as Engine>::GE as PairingGroup>::G2 as DlogGroup>::Affin
...
+/// sufficient on its own; this preserves the wire format and adds the subgroup
+/// check (the SRS `read_ptau` loader applies the same guard).
+fn deserialize_g2_in_subgroup<'de, D, G2>(deserializer: D) -> Result<G2, D::Error>
+where
+  D: serde::Deserializer<'de>,
+  G2: serde::Deserialize<'de> + halo2curves::group::cofactor::CofactorCurveAffine,
+{
+  use halo2curves::group::cofactor::{CofactorCurveAffine, CofactorGroup};
+  use serde::de::Error;
+  let point = G2::deserialize(deserializer)?;
```
abhiramkothapalli, line 64: This implicitly relies on deserialize to manage on_curve checks. It would help to rename deserialize to "checked_deserialize" to better indicate this reliance.

srinathsetty: Isn't that deserialize provided by an external library?

src/provider/bn256_grumpkin.rs
srinathsetty: Other curves do not use custom serialization. Only BN254/Grumpkin are updated.

sai-deng (approved):
Non-blocking scope question: this PR hardens the deserialization / load paths, but invalid affine values can still be created in-memory via public coordinate fields and then passed through APIs like `CommitmentKey::new(...)` or `write_ptau(...)`; they’ll be rejected when reloaded/deserialized, but not necessarily at construction time.

Is that intentionally out of scope for this fix? I think that’s fine if the threat model is untrusted serialized/setup input, but it may be worth documenting that distinction or adding constructor-side validation if these APIs are expected to enforce valid-point invariants too.

MercysJest (approved):
LGTM!

src/provider/hyperkzg.rs
abhiramkothapalli: Sounds good, I was not aware.

abhiramkothapalli:
LGTM

srinathsetty:
> Non-blocking scope question: this PR hardens the deserialization / load paths, but invalid affine values can still be created in-memory via public coordinate fields and then passed through APIs like `CommitmentKey::new(...)` or `write_ptau(...)`; they’ll be rejected when reloaded/deserialized, but not necessarily at construction time.
> 
> Is that intentionally out of scope for this fix? I think that’s fine if the threat model is untrusted serialized/setup input, but it may be worth documenting that distinction or adding constructor-side validation if these APIs are expected to enforce valid-point invariants too.

Good point! Added validation on CommitmentKey::new() path. write_ptau does not need the validation since we already validated incoming paths (added doc).

PR #499, opened 2026-06-19
uniform handling of hyperkzg and mercury

srinathsetty:


sai-deng (approved):

sai-deng:
This now hashes the infinity flag into the HyperKZG transcript, but this path uses `1 = non-infinity` while the RO paths use `1 = infinity`. Should we align the convention?

srinathsetty:
> This now hashes the infinity flag into the HyperKZG transcript, but this path uses `1 = non-infinity` while the RO paths use `1 = infinity`. Should we align the convention?

Good catch! Fixed.

PR #501, opened 2026-07-14
 feat(ppsnark): improve ppsnark memory-check with Logup-GKR

Sun-Jc:
## Summary

This PR improves preprocessing Spartan (ppSNARK), primarily by replacing its inverse-logup memory-check with a Logup-GKR argument. The new default avoids committing to four inverse-oracle polynomials and reduces the final batched PCS opening from 15 polynomials to 11.

The trade-off is proof size: Logup-GKR adds a field-only proof whose size is quadratic in `log N`. In return, the measured end-to-end prover is about **1.3x faster** at medium and large circuit sizes, corresponding to roughly **23-24% less wall time** in the current A/B benchmark.

The original inverse-logup implementation remains available behind the `logup-no-gkr` feature for compatibility, comparison, and deployments that prefer the smaller proof.

## Motivation

The previous ppSNARK memory-check proves, independently for the row and column memories,

```text
sum_i ts[i] / (T[i] + r) = sum_i 1 / (W[i] + r).
```

Its prover materializes four inverse-oracle vectors through two batch inversions, commits to all four vectors, proves four inverse-validity claims in addition to the two balance claims, and includes the four inverse polynomials in the final batched PCS opening. At large `N`, the four length-`N` commitments dominate the memory-check wall time.

Logup-GKR keeps the fractions in projective form and proves their reduction through fractional-add trees. It therefore removes the need to materialize or commit to the inverse polynomials.

### Logup-GKR memory-check

The new memory-check builds four equal-height input layers:

```text
row table:   ts_row / (T_row + r)
row access:       -1 / (W_row + r)
col table:   ts_col / (T_col + r)
col access:       -1 / (W_col + r)
```

Each layer is folded with the projective fractional-add gate

```text
(n_l, d_l) + (n_r, d_r)
  = (n_l * d_r + n_r * d_l, d_l * d_r).
```

The prover and verifier batch the four trees at every layer. The verifier then:

1. verifies the GKR layer reductions;
2. reconstructs the four input fractions from the claimed column evaluations at the shared `eval_point` and reconciles them with the GKR result;
3. checks that the row and column table/access roots each sum to zero; and
4. binds the claimed columns to the committed ppSNARK polynomials through the batched inner sumcheck and final PCS opening.

The GKR reduction adds `O(log N)` layer sumchecks, with a decreasing number of rounds per layer. It contributes `O(log^2 N)` field elements to the proof.

### Opening-point reduction

GKR finishes at its own `eval_point`, while the existing ppSNARK inner sumcheck finishes at `r_inner_batched`. A rerandomization sumcheck carries the seven columns needed by the host reconciliation from `eval_point` to `r_inner_batched`:

```text
L_row, L_col, row, col, ts_row, ts_col, mem_col.
```

This rerandomization claim is included in the existing batched inner sumcheck, so all committed polynomials are still opened once at `r_inner_batched`.

### What is removed and added

Removed from the default path:

- two batch inversions that produce four inverse-oracle vectors;
- four length-`N` inverse-oracle commitments;
- the old six-claim memory sumcheck, including four inverse-validity claims;
- four inverse polynomials and evaluations from the batched PCS input (`15 -> 11` polynomials).

Added to the default path:

- four projective fractional-sum trees;
- one batched GKR argument across the four trees, with `O(log N)` layer sumchecks;
- seven claimed column evaluations at the GKR `eval_point`;
- a rerandomization sumcheck that aligns the GKR final claim with `r_inner_batched`.

The evaluation engine still receives one RLC-batched length-`N` polynomial, so the IPA, HyperKZG, or Mercury evaluation argument itself does not become smaller. The `15 -> 11` reduction saves construction work before `EvaluationEngine::prove`; the main wall-time saving comes from removing the four inverse-oracle commitments.

## Feature gate

Logup-GKR is the default memory-check. The original inverse-logup protocol can be selected at compile time:

```bash
# Default: Logup-GKR
cargo test --release

# Legacy inverse-logup memory-check
cargo test --release --features logup-no-gkr
```

The two configurations have different ppSNARK proof layouts. A proof must be verified with the same feature configuration that produced it.

### Reuse cached multilinear deltas

For a multilinear polynomial split into low and high halves, both round evaluation and binding need the slope

```text
delta = high - low.
```

The evaluation pass now stores `delta` in the high half. After the transcript produces the round challenge, binding reuses it directly:

```text
low <- low + r * cached_delta.
```

This avoids recomputing the subtraction during the bind pass without changing the round polynomial, transcript, or proof. The optimization is applied to:

- the ppSNARK batched inner sumcheck's cubic ABC claim (`L_row * L_col * val`) and its `E` evaluation claim;
- the four-instance Logup-GKR fractional-add layer sumcheck; and
- the multi-column rerandomization implementation when it is used without coefficient fusion.

The default Logup-GKR path fuses the rerandomization columns as described below, so that path binds one combined polynomial instead of seven cached polynomials.

### Fuse the seven rerandomization columns

Once `prove_helper` samples the inner-batch coefficients, the seven rerandomization columns and claims are collapsed into one random linear combination. This reduces the rerandomization state from approximately `7N` field elements to `N` and changes each inner-sumcheck round from seven column scans/binds to one.

The fusion is linear in the same coefficients already used by the batched sumcheck. It preserves the transcript and proof bytes. In the focused A/B measurement, it reduced the inner-batch phase by approximately 25-33% at `2^18` and `2^20` constraints.

### Optimize GKR layer sumchecks

The GKR prover also:

- reuses `EqSumCheckInstance` instead of materializing and rebinding a full equality MLE;
- uses Gruen equality factoring so the equality-polynomial bind is constant time per round;
- uses BDDT claim derivation to compute two `N`-scaling sums instead of three in the common case;
- evaluates the four ppSNARK sub-instances in one parallel pass, hoisting the shared equality factor out of the instance loop;
- computes round data directly and parallelizes only above measured crossover thresholds; and
- uses Horner-style powers for per-instance `lambda` batching.

## Memory management

This PR shortens the lifetime of large prover buffers before later ppSNARK phases allocate their own length-`N` data:

Here, `m` is the constraint count used by the shortened outer sumcheck and `N` is the padded inner-domain size.

- release `poly_Az`, `poly_Bz`, `poly_uCz_E`, and `Cz` after the outer claims are computed (approximately `4m` field elements);
- release the padded relaxed witness after the length-`N` `E` and `W` vectors are materialized (approximately `2m` field elements);
- release the local padded R1CS shape, its lazy sparse-matrix caches, and the assignment `z` after `evaluation_oracles`;
- move, rather than clone, GKR child buffers into each layer sumcheck; and
- consume and drop every GKR tree layer immediately after proving it.

These lifetime changes do not affect the transcript or proof.

## Performance

The following numbers compare the current default Logup-GKR path with `--features logup-no-gkr` using `Bn256EngineKZG` + HyperKZG on an Apple M-series machine with 14 cores. Each row is the median of 10 sequential repetitions after one warm-up.

| Constraints | Logup-GKR | Inverse-logup | Wall-time reduction | Speedup |
|---:|---:|---:|---:|---:|
| `2^16` | 0.667 s | 0.875 s | 23.7% | 1.31x |
| `2^18` | 2.387 s | 3.145 s | 24.1% | 1.32x |
| `2^20` | 8.789 s | 11.405 s | 22.9% | 1.30x |

At `2^20`, the isolated memory-check phase improves from 3.649 s to 0.834 s, or approximately **4.4x**. This phase accounts for essentially the entire end-to-end difference; the two `L` commitments and the final HyperKZG opening are shared costs.

The same `2^20` run measured peak RSS at 9.09 GB for Logup-GKR and 10.92 GB for inverse-logup. RSS remains high because ppSNARK and HyperKZG retain several length-`N` buffers; this PR reduces it but does not claim to solve the full prover-memory problem.

## Proof-size trade-off

Logup-GKR deliberately exchanges proof size for prover time. Let `d = log2(N)`. For four GKR sub-instances, its memory-check-specific payload is approximately

```text
initial roots:                       8 scalars
per-layer split final claims:      16d scalars
compressed layer round polynomials: 3d(d - 1) / 2 scalars
rerandomization claims:              7 scalars
total:                 15 + 16d + 3d(d - 1) / 2 scalars.
```

At `N = 2^20`, this is 905 field elements, or about 28 KiB for 32-byte field elements, before container and serialization metadata. The inverse-logup memory-check instead carries four commitments and four field evaluations.

The shared evaluation argument has the same size in both modes because all PCS inputs are RLC-batched before the evaluation engine is invoked. Users for whom proof bytes, network transfer, or calldata dominate can retain inverse-logup through `logup-no-gkr`.

## Correctness and review notes

- The Logup-GKR core is specialized to ppSNARK's four equal-height row/column table/access instances. It asserts equal `num_vars` and does not add generic padding machinery.
- The verifier defines the transcript order and independently checks every root gate, layer sumcheck, fold-down claim, host reconciliation, and final balance claim.
- Every layer samples a fresh Fiat-Shamir `lambda`; per-instance batching uses distinct Horner powers.
- Performance-only changes preserve the transcript and proof output.
- The default proof format changes because the memory-check protocol changes; `logup-no-gkr` is the explicit legacy fallback.

## Validation

Validated during development under both memory-check configurations:

- `cargo fmt --all -- --check`
- `cargo clippy --all-targets -- -D warnings`
- `cargo clippy --all-targets --features logup-no-gkr -- -D warnings`
- Spartan unit tests, including adversarial Logup-GKR verifier tests
- `cargo test --release test_ivc_nontrivial_with_compression`
- `cargo test --release --features logup-no-gkr test_ivc_nontrivial_with_compression`

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -94,5 +94,8 @@ flamegraph = ["pprof2/flamegraph", "pprof2/criterion"]
 experimental = []
 evm = []
 io = []
+# Use the original inverse-logup memory-check in ppSNARK instead of Logup-GKR.
+# Default (feature off) uses the Logup-GKR memory-check.
+logup-no-gkr = []
 # Enables insecure setup methods that generate random tau. Only for testing!
 test-utils = []
--- a/benches/sumcheckeq.rs
+++ b/benches/sumcheckeq.rs
@@ -4,7 +4,7 @@
 use criterion::*;
 use nova_snark::{
   provider::Bn256EngineKZG,
-  spartan::{ppsnark::MemorySumcheckInstance, SumcheckEngine},
+  spartan::{mem_check_logup::MemorySumcheckInstance, SumcheckEngine},
   traits::Engine,
 };
 use rayon::iter::{IntoParallelIterator, ParallelIterator};
--- a/src/frontend/gadgets/num.rs
+++ b/src/frontend/gadgets/num.rs
@@ -228,7 +228,7 @@ impl<Scalar: PrimeField> AllocatedNum<Scalar> {
             cs.namespace(|| format!("run ending at {i}")),
             &current_run,
           )?);
-          current_run.truncate(0);
+          current_run.clear();
         }
 
         // If `last_run` is true, `a` must be false, or it would
--- a/src/frontend/test_shape_cs.rs
+++ b/src/frontend/test_shape_cs.rs
@@ -141,7 +141,7 @@ where
     let mut s = String::new();
 
     for input in &self.inputs {
-      writeln!(s, "INPUT {}", &input).unwrap()
+      writeln!(s, "INPUT {}", input).unwrap()
     }
 
     let negone = -<E::Scalar>::ONE;
@@ -174,10 +174,10 @@ where
 
         match var.0.get_unchecked() {
           Index::Input(i) => {
-            write!(s, "`I{}`", &self.inputs[i]).unwrap();
+            write!(s, "`I{}`", self.inputs[i]).unwrap();
           }
           Index::Aux(i) => {
-            write!(s, "`A{}`", &self.aux[i]).unwrap();
+            write!(s, "`A{}`", self.aux[i]).unwrap();
           }
         }
       }
--- a/src/spartan/logup_gkr/fraction.rs
+++ b/src/spartan/logup_gkr/fraction.rs
@@ -1,0 +1,95 @@
+//! Projective fraction arithmetic for the Logup-GKR fractional-sum tree.
+//!
+//! A fraction `num/den` is kept in projective form so the GKR circuit never
+//! performs a field inversion: the fraction-add gate combines two children
+//! `(n_l, d_l)` and `(n_r, d_r)` into `(n_l·d_r + n_r·d_l, d_l·d_r)`. This is the
+//! root of why Logup-GKR avoids the inverse-polynomial commitments that the
+//! current ppSNARK memory-check pays for.
+
+use crate::traits::evm_serde::{CustomSerdeTrait, EvmCompatSerde};
+use core::iter::Sum;
+use core::ops::Add;
+use ff::Field;
+use serde::{Deserialize, Serialize};
+use serde_with::serde_as;
+
+/// A projective fraction `numerator / denominator` over the engine scalar field.
+///
+/// Equality of the represented rationals is cross-multiplicative
+/// (`a/b == c/d` iff `a·d == c·b`); this type does not normalize.
+#[serde_as]
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(bound = "F: CustomSerdeTrait")]
+pub struct Fraction<F: Field> {
+  /// Numerator.
+  #[serde_as(as = "EvmCompatSerde")]
+  pub num: F,
+  /// Denominator.
+  #[serde_as(as = "EvmCompatSerde")]
+  pub den: F,
+}
+
+impl<F: Field> Fraction<F> {
+  /// Creates a new projective fraction `num/den`.
+  pub fn new(num: F, den: F) -> Self {
+    Self { num, den }
+  }
+
+  /// The additive identity `0/1` for projective fraction addition.
+  pub fn zero() -> Self {
+    Self {
+      num: F::ZERO,
+      den: F::ONE,
+    }
+  }
+}
+
+/// Projective fraction addition (the 2-to-1 gate): `a/b + c/d = (a·d + c·b)/(b·d)`.
+///
+/// `Fraction` is `Copy`, so `+` takes values with no cost.
+impl<F: Field> Add for Fraction<F> {
+  type Output = Self;
+
+  fn add(self, rhs: Self) -> Self {
+    Self {
+      num: self.num * rhs.den + rhs.num * self.den,
+      den: self.den * rhs.den,
+    }
+  }
+}
+
+/// Sum of a sequence of fractions, folding from the identity `0/1`. Lets a whole
+/// tree level be reduced with `.sum()`.
+impl<F: Field> Sum for Fraction<F> {
+  fn sum<I: Iterator<Item = Self>>(iter: I) -> Self {
+    iter.fold(Self::zero(), |a, b| a + b)
+  }
+}
+
+#[cfg(all(test, feature = "evm"))]
+mod evm_serde_tests {
+  use super::Fraction;
+  use crate::traits::Engine;
+
+  type E = crate::provider::Bn256EngineKZG;
+  type Fr = <E as Engine>::Scalar;
+
+  #[test]
+  fn fraction_has_big_endian_scalar_golden_encoding() {
+    let fraction = Fraction::new(Fr::from(1), Fr::from(2));
+    let config = bincode::config::legacy()
+      .with_big_endian()
+      .with_fixed_int_encoding();
+    let bytes = bincode::serde::encode_to_vec(fraction, config).expect("serialize fraction");
+
+    let mut expected = [0u8; 64];
+    expected[31] = 1;
+    expected[63] = 2;
+    assert_eq!(bytes, expected);
+
+    let (decoded, consumed): (Fraction<Fr>, usize) =
+      bincode::serde::decode_from_slice(&bytes, config).expect("deserialize fraction");
+    assert_eq!(decoded, fraction);
+    assert_eq!(consumed, expected.len());
+  }
+}
--- a/src/spartan/logup_gkr/layer.rs
+++ b/src/spartan/logup_gkr/layer.rs
@@ -1,0 +1,215 @@
+//! Layer type for the Logup-GKR fractional-sum tree.
+//!
+//! One [`Layer`] is a single level of a tree: two multilinear polynomials
+//! (numerator, denominator). The same type serves the input layer, every
+//! internal layer, and the output layer — they differ only in height (the
+//! coefficient length halves each level, `N → N/2 → … → 1`). The same `Layer`
+//! type serves all levels, with no separate input/output variants.
+//!
+//! ppSNARK constructs separate table and access input layers for each of its row
+//! and column relations. Their leaves are `(num, den) = (ts, T+r)` on a table
+//! side and `(-1, W+r)` on an access side.
+//!
+//! # Endianness (critical): MSB-first, pairs are `i` and `i+n`
+//! Nova's MLEs are **MSB-first** (`bind_poly_var_top` folds the top variable by
+//! `split_at(len/2)`). So a level's two children are the halves `x[i]` (top bit
+//! 0) and `x[i+n]` (top bit 1), `n = len/2` — NOT `x[2i]` / `x[2i+1]`. Folding
+//! this way keeps the tree consistent with Nova's sumcheck round order
+//! (`eval_point` challenges bind MSB→LSB), so the GKR point matches the sumcheck
+//! point.
+
+use crate::spartan::logup_gkr::fraction::Fraction;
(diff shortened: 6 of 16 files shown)


Sun-Jc:
## Proof Size Change

- Proof Baseline (total) = `(d+8) G + (3c + 6d + 19) F`
- MemCheck Baseline      = `4 G + 4 F`  (constant, 256 B)
- MemCheck GKR           = `(16d + 1.5*d(d-1) + 15) F`,  `0 G`
- Proof GKR (total)      = shared + GKR memcheck
                         = `(d+4) G + (3c + 22d + 30 + 1.5*d(d-1)) F`
- Δ (GKR − Baseline)     = `-4 G + (16d + 1.5*d(d-1) + 11) F`  (same for total proof and
  memcheck, since the shared skeleton is byte-identical)

## Methodology

- Engine: `Bn256EngineKZG` + HyperKZG PCS (BN254). `test-utils`/`cfg(test)` insecure setup.
- Circuit: `NonTrivialCircuit(num_cons)`, driven through the `DirectSNARK` pipeline
  (`RelaxedR1CSSNARK::prove`). Every measured proof was verified.
- "Proof" = the ppSNARK `RelaxedR1CSSNARK` object (the compressed IVC proof); the
  `DirectSNARK` wrapper adds a constant `1 G + 1 F` on top of both approaches equally.
- **Actual size** = `bincode` (legacy config) serialization length of the real proof (and
  of each individual field), in bytes.
- **Theoretical size** = element counts below; verified to reproduce the measured bytes
  **exactly** (the only delta is `bincode`'s ~8 B length-prefix per `Vec`).
- Measured with a temporary `#[ignore]` in-crate test that mirrors `DirectSNARK::prove`
  and serializes each field; run under `cargo test --release` (GKR) and
  `--features logup-no-gkr` (baseline). The test was removed after measurement (no
  source change is committed for this report).

## Notation & parameters

- `G` = 1 BN254 G1 point (compressed) = **32 B**; `F` = 1 BN254 scalar = **32 B**.
- `c` = # outer-sumcheck rounds = `log2(padded num_cons)`.
- `d` = # inner / mem-check rounds = `log2(N_mem)` (memory-check trace height, in vars).
- `m` = # GKR sub-instances = **4** (`row_table, row_access, col_table, col_access`).
- For the tested `NonTrivialCircuit`: `c = log2(num_cons) + 1`, `d = log2(num_cons) + 2`.

All outer/inner sumcheck round-polys and GKR layer-polys are **cubic** (3 `F` each; the
linear term is compressed out of `CompressedUniPoly`).

## 1. Overall proof layout (`RelaxedR1CSSNARK`)

```json
{
  "comm_L_row":                             "1 G",
  "comm_L_col":                             "1 G",

  "mem_check_proof":                        "<approach-specific — see §2>",

  "sc_outer":                               "3c F        // c cubic rounds",
  "eval_{Az,Bz,Cz,E}_at_r_outer":           "4 F",

  "sc_inner_batched":                       "3d F        // d cubic rounds",
  "eval_{E,L_row,L_col,val_A,val_B,val_C}": "6 F",
  "eval_W":                                 "1 F",
  "eval_{row,ts_row,col,ts_col}":           "4 F",

  "eval_arg (HyperKZG)":                    "(d+2) G + 3d F   // com:(d-1)G, w:3G, v:3d F"
}
```

**Shared (everything except `mem_check_proof`)** = `(d+4) G + (3c + 6d + 15) F`.

## 2. The two `mem_check_proof` layouts

### Baseline — inverse-logup (`logup-no-gkr`) — constant size

```json
{
  "comm_t_plus_r_inv_row": "1 G",
  "comm_w_plus_r_inv_row": "1 G",
  "comm_t_plus_r_inv_col": "1 G",
  "comm_w_plus_r_inv_col": "1 G",
  "eval_t_plus_r_inv_row": "1 F",
  "eval_w_plus_r_inv_row": "1 F",
  "eval_t_plus_r_inv_col": "1 F",
  "eval_w_plus_r_inv_col": "1 F"
  // TOTAL = 4 G + 4 F   (independent of circuit size = 256 B)
}
```

### GKR (default) — grows as O(d^2), no group elements

```json
{
  "proof": {
    "initial_claims": "2m F           // m root fractions (num, den)",
    "final_claims":   "4md F          // d layers x m instances x 2 child fractions",
    "sumchecks":      "3*d(d-1)/2 F   // layer k has k cubic round-polys"
  },
  "rerand_claims":    "7 F",
  // TOTAL = (2m + 4md + 3*d(d-1)/2 + 7) F = (16d + 1.5*d(d-1) + 15) F   (0 G)
}
```

## 3. Actual sizes

`(c, d)` per row = (`c` = outer-sumcheck rounds = `log2(cons_pad)`, `d` = inner/mem-check
rounds = `log2(N_mem)`). Everything in the proof scales with these two log-sizes.

### 3a. Total proof — Baseline vs GKR

| log num_cons | (c, d) | Proof Baseline (FG) | Baseline (KB) | Proof GKR (FG) | GKR (KB) | Δ (KB, ×) |
|---:|:--:|:--|--:|:--|--:|:--|
| 10 | (11,12) | 20 G + 124 F | 4.71 | 16 G + 525 F  | 17.84 | +13.12 (3.8×) |
| 12 | (13,14) | 22 G + 142 F | 5.37 | 18 G + 650 F  | 22.06 | +16.70 (4.1×) |
| 14 | (15,16) | 24 G + 160 F | 6.02 | 20 G + 787 F  | 26.70 | +20.67 (4.4×) |
| 16 | (17,18) | 26 G + 178 F | 6.68 | 22 G + 936 F  | 31.73 | +25.05 (4.8×) |
| 18 | (19,20) | 28 G + 196 F | 7.34 | 24 G + 1097 F | 37.18 | +29.84 (5.1×) |

### 3b. MemCheck payload — Baseline vs GKR (with delta)

| log num_cons | (c, d) | MemCheck Baseline (FG) | Base (KB) | MemCheck GKR (FG) | GKR (KB) | Δ GKR−Base (FG / KB / ×) |
|---:|:--:|:--|--:|:--|--:|:--|
| 10 | (11,12) | 4 G + 4 F | 0.25 | 405 F | 13.38 | −4 G + 401 F / +13.12 / 53.5× |
| 12 | (13,14) | 4 G + 4 F | 0.25 | 512 F | 16.95 | −4 G + 508 F / +16.70 / 67.8× |
| 14 | (15,16) | 4 G + 4 F | 0.25 | 631 F | 20.92 | −4 G + 627 F / +20.67 / 83.7× |
| 16 | (17,18) | 4 G + 4 F | 0.25 | 762 F | 25.30 | −4 G + 758 F / +25.05 / 101.2× |
| 18 | (19,20) | 4 G + 4 F | 0.25 | 905 F | 30.09 | −4 G + 901 F / +29.84 / 120.4× |

The total-proof Δ (KB) equals the MemCheck Δ (KB) exactly — the shared skeleton is
byte-identical, so swapping the memory-check is the *only* difference. The two ratio
columns differ (total 3.8×–5.1× vs memcheck 53×–120×) because the shared part dilutes
the total ratio. GKR drops 4 `G` (the four inverse-oracle commitments) and pays it back
many times over in `F`.

Raw measured bytes (for reference):

| log num_cons | Proof Baseline (B) | Proof GKR (B) | MemCheck Baseline (B) | MemCheck GKR (B) | Witness (num_vars) | Witness size |
|---:|--:|--:|--:|--:|--:|--:|
| 10 | 4824  | 18264 | 256 | 13696 | 2048   | 64 KB    |
| 12 | 5496  | 22592 | 256 | 17352 | 8192   | 256 KB   |
| 14 | 6168  | 27336 | 256 | 21424 | 32768  | 1024 KB  |
| 16 | 6840  | 32496 | 256 | 25912 | 131072 | 4096 KB  |
| 18 | 7512  | 38072 | 256 | 30816 | 524288 | 16384 KB |

Witness size = `num_vars * 32 B` — the raw R1CS witness the ppSNARK compresses.

PR #507, opened 2026-08-19
Add a cryptographic randomness engine to Engine

sai-deng:
## Summary
- add `Engine::RE` as a cryptographically secure, 32-byte-seeded randomness engine
- configure all built-in engines with `ChaCha20Rng`
- use the configured engine for fresh R1CS witness commitment blinds
- release the breaking trait update as `nova-snark` 0.75.0

## Why
Randomness is a cryptographic engine choice alongside the transcript and commitment engines. An associated type lets generic protocols initialize a secure deterministic stream without hard-coding a concrete RNG, while ordinary witness construction continues to seed from fresh system entropy.

## Scope
This PR applies the configured engine to fresh R1CS witness commitment blinds. Existing protocol-specific random sampling paths remain unchanged and can migrate separately when they need caller-seeded control.

## API
```rust
pub trait Engine {
  type RE: RngCore + CryptoRng + SeedableRng<Seed = [u8; 32]> + Send;
  // ...
}
```

## Validation
- all-target compilation with `experimental`
- strict Clippy with `experimental`
- deterministic `Engine::RE` seed-reproducibility test

### DIFF
--- a/src/r1cs/mod.rs
+++ b/src/r1cs/mod.rs
@@ -79,6 +79,8 @@ pub struct R1CSWitnessBlind<E: Engine>(E::Scalar);
 impl<E: Engine> R1CSWitnessBlind<E> {
   /// Samples a fresh witness commitment blinding factor.
   pub fn random() -> Self {
+    // Seed the configured engine from system entropy so all Nova randomness
+    // follows the same engine policy, including caller-seeded protocols.
     Self(E::Scalar::random(&mut E::RE::from_entropy()))
   }
 
--- a/src/r1cs/mod.rs
+++ b/src/r1cs/mod.rs
@@ -17,7 +17,7 @@ use crate::{
 use core::cmp::max;
 use ff::Field;
 use once_cell::sync::OnceCell;
-use rand_core::OsRng;
+use rand_core::{OsRng, SeedableRng};
 use rayon::prelude::*;
 use serde::{Deserialize, Serialize};
 use serde_with::serde_as;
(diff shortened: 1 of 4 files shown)


src/r1cs/mod.rs
```diff
@@ -79,7 +79,9 @@ pub struct R1CSWitnessBlind<E: Engine>(E::Scalar);
 impl<E: Engine> R1CSWitnessBlind<E> {
   /// Samples a fresh witness commitment blinding factor.
   pub fn random() -> Self {
-    Self(E::Scalar::random(&mut OsRng))
+    // Seed the configured engine from system entropy so all Nova randomness
```
srinathsetty, line 82: nit: is this comment necessary?

src/r1cs/mod.rs
```diff
@@ -79,7 +79,9 @@ pub struct R1CSWitnessBlind<E: Engine>(E::Scalar);
...
+    // follows the same engine policy, including caller-seeded protocols.
+    Self(E::Scalar::random(&mut E::RE::from_entropy()))
   }
 
   /// Wraps a protocol-derived blinding factor.
```
srinathsetty, line 87: Given E::RE, do we still need this protocol-derived blinding factor path?

sai-deng:
Closing in favor of the smaller HyperNova API: proof methods now accept a caller-scoped generic `R: RngCore + CryptoRng`, so no Nova `Engine::RE` change or release is required. The engine-associated RNG idea can be reconsidered separately if Nova later adopts a library-wide randomness policy.

